PR #2, opened 2017-04-07
development/astyle: Updated for version 3.0

N0NB:


willysr:
merged in my branch 

closed #2

N0NB:
* On 2017 07 Apr 20:32 -0500, Willy Sudiarto Raharjo wrote:
> merged in my branch

Thanks, Willy.

Will using the GitHub pull request notifications be sufficient or should
I CC you or the mailing list as well?

- Nate

-- 

"The optimist proclaims that we live in the best of all
possible worlds.  The pessimist fears this is true."

Ham radio, Linux, bikes, and more: http://www.n0nb.us

willysr:
I think github PR should be enough. I didn't set notification before so i wasn't notified, but it should be set now

PR #3, opened 2017-06-06
development/hhvm: Updated for version 3.19.2

belka-ew:


willysr:
Applied in https://git.slackbuilds.org/slackbuilds/commit/?id=004b231ba

Closed #3

PR #4, opened 2017-06-11
Updates ctags, hhvm 3.20.2

belka-ew:
I just realized that I haven't updated hhvm for a long time :)

PS: Is it ok if multiple updates are in a single pull request (of cause in separate commits)?

Thanks.


willysr:
Better to have different PR so it's easier to merge for me :)

willysr:
merged in my branch

thanks

PR #12, opened 2017-08-04
development/hhvm: Updated for version 3.21.0

belka-ew:


willysr:
merged in my branch

thanks

PR #14, opened 2017-08-10
libraries/luadbi: Fixed download link

aclemons:
Homepage remains unchanged. Seems there is not a canonical replacement
at the moment.


willysr:
merged in my branch

thanks

PR #23, opened 2017-09-08
development/dmd: Updated for version 2.076.0

belka-ew:
Sorry, I should fix my latest dmd commit. Lately I made the package a bit smaller, but broke some imports. This should fix it.
I do a commit against master. If I should rebase it on you slackbuilds' branch, I can do that.
And I think that I will switch to building dmd from source soon since I know now how to get it working.


willysr:
merged in my branch

PR #31, opened 2017-11-02
Scattering of URL and MD5SUM fixups

aclemons:
A set of fixups I've found where URLs are no longer valid or the contents / MD5SUMs have changed + a couple of patches for builds broken by curl updates.


willysr:
merged in my branch

PR #39, opened 2017-11-18
Fix syntax error for mapnik

AMDmi3:


willysr:
I have this patched in my branch already

PR #46, opened 2017-12-16
Add missing quote

AMDmi3:


willysr:
I have this fixed on my branch

willysr:
the public update will happen in the next 2 days

willysr:
i have pushed the public update just now

AMDmi3:
Thank you!

PR #47, opened 2017-12-29
eclipse updates

aclemons:
I submitted these before the public update on 21 December and they have now disappeared from the pending queue, but were not applied. Now when I try to submit them through the submission form on slackbuilds.org, they are rejected because a submission is supposedly already pending approval.


rworkman:
Merged into my branch. Sorry about that - no idea what happened to them before. Given that I've not done a public update in a *long* time, I strongly suspect that I screwed up somewhere.

PR #56, opened 2018-01-12
development/hhvm: Updated for version 3.23.4

belka-ew:
And the last minute hhvm update to the version released today.


willysr:
merged

PR #94, opened 2018-04-09
Fix LUA_ROOT path

N0NB:
The build script omitted a trailing '/' when removing "local" from the LUA_ROOT definition.  This caused Hamlib bindings for Lua to be installed in "/usrlib*/ instead of "/usr/lib*".


willysr:
merged

PR #108, opened 2018-05-09
network/postfix: Retain license files

Slamdunk:
`/etc/postfix/{TLS_,}LICENSE` files are part of post-install scripts:

https://github.com/vdukhovni/postfix/blob/v3.3.0/postfix/conf/postfix-files#L143-L144

Removing them gives me the error:
`chown: cannot access '/etc/postfix/LICENSE': No such file or directory`

In the file:
https://github.com/vdukhovni/postfix/blob/v3.3.0/postfix/conf/post-install#L570

Retaining them works.

### DIFF
--- a/network/postfix/postfix.SlackBuild
+++ b/network/postfix/postfix.SlackBuild
@@ -172,8 +172,6 @@ make non-interactive-package \
   manpage_directory=/usr/man \
   readme_directory=/usr/doc/$PRGNAM-$VERSION/README_FILES
 
-rm -f $PKG/etc/postfix/{TLS_,}LICENSE
-
 install -D -m 0644 -oroot -groot $CWD/rc.postfix $PKG/etc/rc.d/rc.postfix.new
 find $PKG/etc/postfix -type f ! -name "*.default" ! -name "*.proto" ! -name "*.out" \
   ! -name "postfix-files" -exec mv {} {}.new \;


willysr:
Hi

i tested the build without this patch and it seems to work normally here.

Slamdunk:
Hi, if `/etc/postfix/{TLS_,}LICENSE` are already present it works, but on a fresh new installation it doesnt.
This is how I reproduce the bug:

1. Grab the ISO from https://mirrors.slackware.com/slackware/slackware-iso/slackware64-14.2-iso/
1. Install a fresh copy selecting only A/D/K/L/N sets
1. Login as root
1. `wget --content-disposition "https://slackbuilds.org/slackbuilds/14.2/network/postfix.tar.gz"`
1. `tar -xvf postfix.tar.gz`
1. `cd postfix`
1. `wget --content-disposition "http://cdn.postfix.johnriley.me/mirrors/postfix-release/official/postfix-3.3.0.tar.gz"`
1. `groupadd -g 200 postfix`
1. `useradd -u 200 -d /dev/null -s /bin/false -g postfix postfix`
1. `groupadd -g 201 postdrop`
1. `./postfix.SlackBuild`
1. `installpkg /tmp/postfix-3.3.0-x86_64-1_SBo.tgz`

#### Last command output
```
Verifying package postfix-3.3.0-x86_64-1_SBo.tgz.
Installing package postfix-3.3.0-x86_64-1_SBo.tgz:
PACKAGE DESCRIPTION:
# [...]
Executing install script for postfix-3.3.0-x86_64-1_SBo.tgz.
chown: cannot access '/etc/postfix/LICENSE': No such file or directory
Package postfix-3.3.0-x86_64-1_SBo.tgz installed.
```

It says is installed and it is, but the post-install script exited with the error and didn't ended entirely.

willysr:
fixed in my branch

Slamdunk:
Hi, may I ask what that means?
I don't see the commit in the repo and I'm not aware of your release process :)

belka-ew:
https://git.slackbuilds.org/slackbuilds/commit/?h=willysr

Slamdunk:
I see, thanks for the link.
The next time I hope the original commit can be cherry-picked so the real author will be maintained.

willysr:
Sorry about that
It should be fixed now: https://git.slackbuilds.org/slackbuilds/commit/?id=e8f5a4b79

Slamdunk:
Thank you for your patience :+1:

PR #150, opened 2018-09-13
development/gcc-d: Updated for version 2.081.2

belka-ew:
It is a big update. I'll describe it briefly since the build script changed a lot.

- D is updated to 2.081.2 which is a D frontend written in D itself, only supported by GCC 7 and GCC 8
- The package ships GCC 8 (C, C++ and D)
- First GCC 7 is built with an older D frontend in C++. Then it is used to bootstrap GCC 8, which is finally installed
- The compiler is installed under /opt/gdc with man pages and info files under /usr and symlinks under /usr/bin (symlinks like gcc-8, g++-8 and also gdc-8 and gdc - so they don't overwrite anything from GCC 5)
- I commented out the code I used for installation under /usr since I can reuse it later on -current which has a newer compiler. But if these are too many comments, I'll remove them
- The build script depends now on 2 GCC (7 and 8) and 2 GDC versions (one in C++ for bootstrapping and the final D version)
- I've updated the README appropriately
- I have been using the package for a week now and haven't noticed any problems

### DIFF
--- a/development/gcc-d/README
+++ b/development/gcc-d/README
@@ -1,15 +1,12 @@
 This is GDC (GNU D Compiler), the D frontend for GCC.
 
-D is a general-purpose programming language with static typing,
-systems-level access, and C-like syntax. It combines efficiency,
-control and modeling power with safety and programmer productivity.
+D is a general-purpose programming language with static typing, systems-level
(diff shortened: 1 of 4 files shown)


willysr:
merged

PR #207, opened 2018-12-31
Updates to 6 of my packages

bassmadrigal:
I have 6 version updates in this PR. If you need them separated out (they already each have their own commit), I can submit separate PRs for each.

NOTE: development/cloc does require a new dependency (perl-parallel-forkmanager) that has already been submitted as a new submission on the SBo site.


willysr:
merged

PR #223, opened 2019-01-22
network/discord: Update to 0.0.8

bassmadrigal:
This is an update to discord v0.0.8. I did have a single report of this not working on -current, but I don't run that and can't test it. If you end up being aware of any -current issues, let me know and I'll see what I can do.


willysr:
merged

PR #250, opened 2019-02-28
update to 3.0.4, thanks to Tim for the qt5 patch

notklaatu:


willysr:
All requirements should go into REQUIRES, not in README

notklaatu:
I'm not clear how to deal with the jack-audio-connection-kit vs. jack2 dependency.

Either one satisfies the requirement. Should I list both in REQUIRES? should I list both in README and place %README% where JACK would normally be listed?

belka-ew:
@notklaatu pick one as the "default dependency" and put it into REQUIRES. And mention in the README that you can also use another one instead. Java packages do the same: jre, jdk or openjdk can satisfy the requirements.

belka-ew:
Also the commit message should be "multimedia/MuseScore: Updated for version 3.0.4"

And I see now that the version in the SlackBuild (3.0.3) doesn't match the version in the .info file (3.0.4).

notklaatu:
Updated.

belka-ew:
@willysr ping.

willysr:
Hi @notklaatu 

can you squash this two commit into a single commit?

Sorry for late reply, i was on vacation and i just got back

notklaatu:
Rebased into one commit

PR #265, opened 2019-03-19
system/qtfm: Updated for version 6.1.8.

rodlie:
Note that this is the last version that will build with Qt 4. Version 6.2+ will require Qt 5.


willysr:
merged

PR #267, opened 2019-03-22
development/rtags: Updated for version 2.30.

maramon:
I'm aware that even newer version 2.31 is available. I didn't update to that version, because it contains a bug which I don't want to pull in.


willysr:
merged

PR #328, opened 2019-07-04
system/ansible: Updated for version 2.8.2, switch to python3.

majekw:
- update to 2.8.2
- switch to python3
- update dependencies (paramiko is no longer hard requirement)
- add note about additional dependencies required at run time (for example some modules may still require paramiko)


willysr:
merged

PR #337, opened 2019-07-22
python/python3-lazy-object-proxy: Fixed dependencies.

maramon:
If dependency setuptools-scm was not present, the SlackBuild tried to download it.


willysr:
merged

PR #347, opened 2019-08-13
Flare

Ythogtha:
Is it okay to merge request with two commits, on two packages ?
They are totally related and one goes with the other one.


willysr:
merged

PR #348, opened 2019-08-13
system/dosbox-dev: Updated for revision 0.74.3.r4253.

Ythogtha:
I changed version from 0.74.r<REVISION> to 0.74.3.r<REVISION> because it is a development version ahead of 0.74.3 release, and repology thinks this version is outdated even though it's the latest one available.
Anyhow, since they are making new relases more often than one in eight years, the dosbox-dev build is less important, so basically all that doesn't matter very much.


willysr:
merged

PR #396, opened 2019-10-20
misc/protobuf3: Fix MD5SUM.

aclemons:
This was not updated in 6b96212dea to match the m5sum of the new
version's tarball.

Signed-off-by: Andrew Clemons <andrew.clemons@gmail.com>


willysr:
Fixed in my branch already

PR #416, opened 2019-11-11
Fix syntax

AMDmi3:


willysr:
It's fixed in Chris' branch already

AMDmi3:
I wonder if it's not in the master, as it breaks repository parsing.

willysr:
It will be part of master in the next public update, which is usually happening in the weekend

PR #423, opened 2019-11-18
Updates for 6 SlackBuilds

bassmadrigal:
development/cloc: Update to 1.84
development/meson: Update to 0.52.0
libraries/libmediainfo: Update to 19.09
multimedia/mediainfo: Update to 19.09
network/openresolv: Update to 3.9.2
network/transgui: Update to 5.18.0


willysr:
please keep the permission stays in 644 instead of 755

bassmadrigal:
Oops, I didn't realize I had changed them. Not sure how that happened...

Want me to resubmit with corrected permissions? If so, can I do that through this same pull request (I'm still pretty new to pull requests).

willysr:
yes, please
keep the permission 

you can use this git hooks (save it as pre-commit in .git/hooks):
```
ERROR="no"
FILES=$(git diff-index --cached --name-only HEAD)
HAS_INFO=$(git diff-index --cached --name-only HEAD | grep "\.info")

if ! [ -z "$FILES" ]; then
  for i in $FILES; do
    if [ -e $i ]; then
      MODE=$(stat -c "%a" "$i")
      if ! [ $MODE = "644" ]; then
        echo "$i has wrong mode set ($MODE instead of 644)"
        ERROR="yes"
      fi
    fi
  done
fi

if [ "$ERROR" = "yes" ]; then
  echo "Exit"
  exit 1
fi
```

PR #436, opened 2019-12-27
Passwdqc 1.4.0

pyllyukko:


pyllyukko:
Hmm. For some reason there was a change in ChangeLog also. Will create a new PR.

pyllyukko:
#437

PR #442, opened 2020-01-05
Fix proj info file syntax

AMDmi3:


willysr:
It has been fixed in one of the branch

PR #462, opened 2020-01-21
python/pycryptodomex: Renamed from python/pycryptodome.

maramon:
The new name is consistent with naming in PyPI.


willysr:
please use normal submission form as it must be populated by the database.

maramon:
Submitted via submission form.

PR #473, opened 2020-02-06
libraries/htmlcxx: new maintainer

Ythogtha:


willysr:
this one also doesn't apply against latest master

PR #480, opened 2020-02-10
desktop/x-tile: updated for version 2.6

Ythogtha:


willysr:
i made some changes here, so make sure to grab the latest changes after public update

PR #514, opened 2020-04-16
games/ppsspp: Update for ppsspp 1.9.4-1

LinRs:
- add patch for libpng-1.6
- enable sdl2 interface
- add option for build with qt5


willysr:
merged
you missed python3

PR #520, opened 2020-04-21
network/keepalived: Updated for version 2.0.20.

majekw:
- update to newest version
- disabled nftables support as it doesn't compile on 14.2
- removed dependency of optional sysvinit-functions and simplified rc.keepalived
- moved daemon configuration to /etc/default


willysr:
merged

PR #527, opened 2020-05-03
Update python3-astroid and python3-pylint

maramon:
These updates are in the same pull request because the new pylint has a version specific dependency to the new astroid.


willysr:
merged

PR #578, opened 2020-09-02
Glewlwyd

Ythogtha:
That's OK if I group commits pertaining to the same software ?
All commits are updates of Glewlwyd's dependencies.


willysr:
merged
yes, it's fine

PR #607, opened 2020-10-10
development/cargo-vendor: Removed (part of cargo).

aclemons:
Slackware 14.2 was bumped to rust 1.46.0 whose cargo now
includes the vendor subcommand.


willysr:
merged

PR #608, opened 2020-10-12
Updates for stuntrally and fix ogre download

bassmadrigal:
I switched stuntrally to the binary release until the port to ogre-1.11+ is complete (or the PR leads to a functional build, right now it segfaults). I tested it and it seems to work without any requiring any additional dependencies.

I also noticed the ogre download for the version on SBo right now is no longer works. I switched to a github download (should be the same release) and changed the SlackBuild to work with the new name. If you end up merging the new bullet stuff (including the update to a newer ogre), this won't be needed, but if you're not ready to push those changes, you may want to incorporate this.


willysr:
@bassmadrigal did you test using this changes?
https://git.slackbuilds.org/slackbuilds/log/?h=bullet-review

bassmadrigal:
I sent a companion email, but I suppose I should've included it here as well.

I could not get stuntrally to run with the new ogre (segfault when you launch it), so I switched it to the binary release until I can find a fix. stuntrally can be pushed immediately and isn't affected by the bullet-review branch.

The ogre patch is in case you don't push the newer version before the next public update. The v1.9 download is a 404, so I switched it to the github release and adjusted the SlackBuild for the new name.

There were a few issues with the ogre SlackBuild in the bullet-review branch that I mentioned in the email to you. One, pugixml is missing from the REQUIRES and the cmake configure process will download imgui. I fixed both those and included a patch in the email to apply to that branch.

willysr:
ok, it's now pushed on https://git.slackbuilds.org/slackbuilds/log/?h=bullet-review
please make sure i applied it correctly

willysr (commented):
Did you mean to write lib64 instead of lib/64 ?

bassmadrigal:
It's supposed to be lib/64. Their binary version supports both 32 and 64 bit, and have things in lib/{32,64} respectively. Those is why I used /opt instead of /usr as they were using non-standard locations for a lot of files.

If you have a clean 32bit build, would you mind testing to see if /opt/stuntrally/lib/32/libstdc++.so.6 causes a segfault like the 64bit version?

willysr:
i changed it to {32,64} so it will remove both 
i don't have 32 build machine :(

bassmadrigal:
That works. I don't have a 32bit machine either, so I was honest 6 going to rely on bug reports to fix it if reported. Probably safe to remove it on both arches.

PR #611, opened 2020-10-19
Update to 3.12.1-rc1

notklaatu:


willysr:
hi @notklaatu 

can you fix some issue with this PR?
- The version should not have "-". Probably best to change it into "_"
- The VERSION in .info doesn't match with the real version which is rc1
- The README has been fixed by Urchlay in last public update. Please wrap them.

notklaatu:
>     * The version should not have "-". Probably best to change it into "_"

Do you mean change it in the makepkg command at the end, or throughout the entire SlackBuild script?

willysr:
in the VERSION part

willysr:
you can force-pushed it, so it stays in this PR

PR #617, opened 2020-10-30
Update to version 3.1.12-rc1

notklaatu:
Updated to version 3.1.12-rc1 as 3.1.12_rc1


willysr:
Hi @notklaatu 

please use the updated README fixed by Urchlay (your local copy is outdated)
for the tarball, let's use the same format as before so you don't have to make changes in the SlackBuild as well
the only change needed is the VERSION in both .info and .SlackBuild to use _ instead of -

willysr:
closing as no response

PR #655, opened 2020-12-29
Initial slackbuild for prometheus 2.23.0

earies:


willysr:
Please use normal submission form for new script. PR/MR is only for updating scripts

PR #673, opened 2021-01-13
Update Ledger to the latest 3.2.1 version

sanel:
Updating necessary links, but also adding small patch that will allow
compiling it with CMake 3.5, which is standard on Slackware 14.2.

With CMake versions less than 3.7, it will compile Ledger without
python support, because it will skip all CMake python detection
checks. For versions >= 3.7, it will proceed with detecting necessary
python dependencies.


willysr:
merged

PR #679, opened 2021-01-27
network/enet: Updated for version 1.3.17 + new maintainer

sbolokanov:
Taking over enet, if that's ok...
I've forwarded you e-mail from the original maintainer.


willysr:
merged

PR #690, opened 2021-02-04
network/ubertooth: Updated for version 2020-12-R1.

pyllyukko:
Also change the version format to be compatible with [Repology](https://repology.org/project/ubertooth/versions).


willysr:
merged

PR #694, opened 2021-02-05
perl/perl-Cairo-GObject: Update to version 1.005

mab974:


willysr:
hi @mab974 please make sure the commits are linear so they can be easily merged into SBo branches (and soon to master). I would sugges to make a new PR instead since it would be easier.

PR #705, opened 2021-02-16
perl/perl-File-Remove: Updated for version 1.60. 

mab974:


willysr:
please fix this PR as well

PR #712, opened 2021-02-22
Update python3-astroid and python3-pylint

maramon:
These commits are in the same merge request because the new version of python3-pylint requires the new version of python3-astroid.


willysr:
merged

PR #724, opened 2021-02-27
games/ags: updated for version 3.5.0.30

Ythogtha:
This is a big rework of the slackbuild.
I inherited it as a complex build with its own version of allegro4 and dumb, all installed inside /opt/ags/.
I decided to use Sbo's allegro4 and add a slackbuild for dumb library.
It makes for a cleaner build, and easier maintenance of it's dependencies.

Also, this PR shouldn't be merged before the new library/dumb slackbuild is accepted from the submit form on slackbuilds.org :)


willysr:
merged

PR #730, opened 2021-03-06
Upgrade python3-astroid and python3-pylint

maramon:
These updates are in the same pull request because the new version of python3-pylint requires the new version of python3-astroid.


willysr:
merged

PR #732, opened 2021-03-11
libraries/dumb: build both static and shared libraries

Ythogtha:
Building both static and shared libs.
Static are needed to build ags, which is the reason for this slackbuild.
But allegro autodetects the presence of libdumb and tries using it, which will fail if there are only the static libs available!
Since dumb is quite small, it's not a big deal to build it twice for shared and static.


willysr:
merged

PR #733, opened 2021-03-11
games/endless-sky: allows for building on both -current and -14.2

Ythogtha:
Update only for building on -current.
If you feel there's no need, I can postpone the change until next version of endless-sky, as this won't change anything on -14.2.
But as 15.0 is drawing near, it might be worth doing.


willysr:
merged

PR #751, opened 2021-03-24
perl/nqp: Updated for version 2021.03

mab974:


willysr:
merged
please don't add extra white space at the end onf .info file

PR #753, opened 2021-03-24
4 Updates

bassmadrigal:
3 version bumps (development/meson and network/{discord,openresolv}) and adding a missing dependency to the REQUIRES of lirc.


willysr:
merged

PR #767, opened 2021-04-04
games/freeciv: wrong md5sum

Ythogtha:
Sorry, race condition in my brain : prepared the PR before having downloaded the source...
Brain procedures are under review, due to being reworked soon ;)


willysr:
merged

PR #768, opened 2021-04-05
games/warzone2100: updated for version 4.0.0

Ythogtha:


Ythogtha:
Argh, does not build on 14.2, because sqlite is too old...

PR #774, opened 2021-04-12
libraries/rhonabwy: updated for version 0.9.999

Ythogtha:


Ythogtha:
-Wno-error=maybe-uninitialized is there to allow building in -current, and it does not affect 14.2.

willysr:
merged

PR #783, opened 2021-04-19
Update of zoom-linux slackbuild to 5.6.16775.0418

earies:


willysr:
Hi @earies 

since we are now in development cycle and zoom is updated quite frequent, would you mind hold updating this script ?
this changes will only goes into master and not to 14.2 branch, so it will not be pushed to our website until 15.0 repository is out.

PR #787, opened 2021-04-22
Yth requirements for 15.0

Ythogtha:
A bunch of commits removing requirements which are now parts of slackware-15.0-beta.
python3, qt5, SDL2, OpenAL, ffmpeg and jansson.
Also removed liboauth from lgogdownloader because it's apparently not really required, I must have missed that in the past.


Ythogtha:
Ok, this has already been done, sorry....

rworkman:
No worries - trying to keep up with git master is difficult for us too right now.
Once we get the "low hanging fruit" finished, it will settle down a bit and be easier to do these sort of merge requests.
Much appreciation for your work here.

PR #789, opened 2021-04-22
system/sqldiff: updated for version 3.35.4

Ythogtha:
That one is special : sqldiff follows exactly the sqlite version in Slackware, and I'm building it with the same options that are used to build sqlite, hence the big changes in the .Slackbuilds, they come from stock -current, also download URL from mirrors.slackware.com : we use exactly the same sources as the one that are included in Slackware.
So if sqlite does not change until 15.0, it may well be the only commit there until 15.1.


willysr:
i upgraded to 3.55.5 since the source has changed already

Ythogtha:
Looks lie -current is still a moving target, thanks ^^

PR #791, opened 2021-04-22
games/warzone2100: updated for version 4.0.1

Ythogtha:
Because 4.0.0 is less than 3 weeks old, and was the first version for which slackware-14.2 was too old, so it's worth bumping it.


willysr:
merged

PR #792, opened 2021-04-22
libraries/allegro: Removing workaround for 14.2, and updated for version 5.2.7.0

Ythogtha:
Because I had to include an ugly workaround only a few weeks ago, and now it serves no purpose, so cleaning the slackbuild and bumping it into the future !


willysr:
merged

PR #795, opened 2021-04-22
games/lgogdownloader: removed liboauth requirement

Ythogtha:
Because liboauth isn't used.
I don't know when that happened, I must have missed it then.


willysr:
merged

PR #798, opened 2021-04-27
development/hopper: Updated for version 4.7.4.

pyllyukko:


willysr:
please note that PR is now must be tested against -current, so qt5 is no longer needed as it's included already

pyllyukko:
Ok. I missed the announcement on that. I'll make another PR.

PR #808, opened 2021-05-01
Python buildbot worker

Ythogtha:
I've acquired the authorization from Markus Reichelt <slackbuilds@mareichelt.de> to take over his very old and outdated buildbot-slave slackbuild.
This requires to rework all its requirements, also quite old, and mostly python2.

As I'd like to have my python slackbuilds to be python3 only (except when it is explicitly needed) for slackware 15.0, and have a consistent naming pattern, I've decided to submit all those dependencies as "python-XXX" slackbuilds, probably condemning the old ones to be removed.

This draft is for testing purposes as new slackbuilds needs to be submitted through the form, and we should submit new slackbuilds when 15.0 is out.
But I'd like any remark about that, if necessary, before going forward.

I still don't have explicit authorization from Markus to take over python-twisted, and I won't have authorization from Mikko Värri since his email address is invalid and he didn't submit anything for more than five years.

However, this work is not in progress : it builds and runs fine as it is.
The only remaining thing to do is to do things by the book :)
I'm also working on python-buildbot, but still waiting on a reply from David Spencer.


willysr:
I think this is suitable when 15.0 repo is open again as this is a WIP status, where PR should be a finished one

PR #845, opened 2021-06-12
Refactored and upgraded vscode-bin

partizansoft:
Refactored and upgraded vscode-bin to version 1.57.0. Since I use this IDE from time to time, I could take maintenance on this package.


willysr:
I updated to 1.57, but i'm keeping the old style of my own SlackBuilds and maintainership

PR #848, opened 2021-06-12
libraries/imlib2_loaders: Updated to version 1.7.1

partizansoft:
It seems that original author abandoned this package, so I would take it for maintaining


willysr:
merged

PR #855, opened 2021-06-14
office/evince: Updated to version 3.39.1

partizansoft:
This is not my script, so I think the author should review it. The reason for the update: older version had terrible bug with marking text in pdf files


willysr:
Maintainer is active and he will update it when 15.0 gets released, so i will wait until that happened.

PR #885, opened 2021-07-11
libraries/gspell: Added library which is dependency for some GNOME apps

partizandevel:
This library is required for spell checking by some GNOME applications. Newer evince is one of them.


willysr:
merged

PR #920, opened 2021-08-09
office/pdfpc: fix compilation against current vala

ArTourter:
due to an API change in vala 0.52.x, compilation was broken. It seems
that vala will be fixed in a future release (for 0.52 and 0.54), for the
moment a patch is required as discussed in
https://github.com/pdfpc/pdfpc/issues/594

### DIFF
--- a/office/pdfpc/pdfpc.SlackBuild
+++ b/office/pdfpc/pdfpc.SlackBuild
@@ -27,7 +27,7 @@ cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=pdfpc
 VERSION=${VERSION:-4.5.0}
-BUILD=${BUILD:-1}
+BUILD=${BUILD:-2}
 TAG=${TAG:-_SBo}
 PKGTYPE=${PKGTYPE:-tgz}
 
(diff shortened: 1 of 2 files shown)


fnevgeny:
FYI, this fixes the compilation, but videos will crash pdfpc. You should instead apply the Vala fix.

ArTourter:
Thanks for the note @fnevgeny , I had not realised the caveat.

Since vala is part of the distribution, we will probably have to wait for upstream to officially release a version with the fix which hopefully will not be long. In the meantime, I leave it slackbuilds.org admins to decide if it is worth having a package that compiles but that is not fully functional, or one that doesn't compile at all.

fnevgeny:
> Since vala is part of the distribution, we will probably have to wait for upstream to officially release a version with the fix which hopefully will not be long.

It isn't: https://gitlab.gnome.org/GNOME/vala/-/commit/873c879367d1a4d7265e32dda55d4c01d5dd957b

How applying that (real) patch is principally different from (cosmetically) patching pdfpc?

willysr:
As for vala, it should be directed to Patrick since it's part of the core Slackware packages.
as for pdfpc, we still have time until 15.0 is ready and by that time, it should have been fixed.
You can keep this issue open until it's solved later.

willysr:
@ArTourter New Vala is now pushed to main Slackware. Can you test whether it fixed the issue with pdfpc?

ArTourter:
Vala 0.52.5 which include the fix has now been pushed to core slackware-current. I have tested the original slackbuild of pdfpc and it compiles and works properly.
This patch is therefore no longer required so I am closing this.

PR #925, opened 2021-08-18
Insync update

cduston44:
Fixed broken link in .info.


willysr:
I think this patch is not rebased on latest master since what we have in our repo is already at the same version with this patch applied

cduston44:
Apologies, I am still learning the ins and outs of git. I will try again on a fresh fork, perhaps.

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #927, opened 2021-08-21
Upgrade python3-astroid and python3-pylint

maramon:
These commits are in the same pull request because the version dependencies between astroid and pylint require them to be upgraded in lockstep.


willysr:
merged

PR #990, opened 2021-10-23
Fix Patch

RSKYS:
Hi,
These are several packages need to get upgraded as they have some issues and I'm experiencing them;
This list might updated.
pulsectl: Issues with live connecting track files after fresh mount.

Regards, Pouria


willysr:
merged

PR #1015, opened 2021-10-30
audio/tauonmb: Fix for version 6.8.0.

RSKYS:
Hi Willy,
Sorry but seems I've missed the compiler and it's locale;
Here's the fix + polish a bit..

Regards,
Pouria.


willysr:
merged

PR #1018, opened 2021-11-01
discord: Version bump to 0.0.16.

bassmadrigal:


RSKYS:
Glad we're here,
@bassmadrigal Please check [this request](https://github.com/bassmadrigal/SBo-forked/pull/3) before this merge.

Regards,
Pouria.

willysr:
I have merge previous PR by @RSKYS here: https://git.slackbuilds.org/slackbuilds/commit/?h=willysr&id=0a6af2f09889e83ca85832d88e12a20560ab99f0

can you review the additional changes @bassmadrigal ?

RSKYS:
Although this still hasn't really fixed my issue with Presence;
Offered these changes along all of my modifications anyway.

Almost every dependency resolved by Slackware, everything should've worked;
I strongly think this is python3.10 to blame in case of Presence.

bassmadrigal:
I don't currently see a reason to change the install location to /opt/, however, if there's a benefit behind it, I'm open to it.

It doesn't seem to be more standard in /opt/ since the current package seems to be installed properly into /usr/ and doesn't all reside under a single directory. If I was using the tar.gz download from discord, that would make a lot more sense as everything resides in a parent discord/ folder.

RSKYS:
I was testing this to see if it works with binary on certain locations;
Other thing I've tried was at /usr/lib64 just like Fedora standard packaging. You can take the merge or revert it back eventually, this is just suggestion.

Regards,
Pouria.

willysr:
i'll take @bassmadrigal patch since he's the maintainer for this script

RSKYS:
Sure.

PR #1025, opened 2021-11-04
gis/cartopy: update dependency in info file

ArTourter:


ArTourter:
Note I didn't update the package number as the dependency are as far as I can see runtime requirements so recreating the package should not be necessary.  Without pyproj installed, cartopy doesn't work, and although gdal is not as such required, the functionality it offers is, justifies its inclusion. if you prefer the package number to be incremented, then feel free.

willysr:
merged

PR #1030, opened 2021-11-05
New

RSKYS:
Hi Willy,
These are my SBo plans for Slackware 15.
I'm being patient due "SBo: Please be patient..",
I'm gonna add some other more builders in another commits later;
I have some couple mores too actually.

Regards,
Pouria

PR #1031, opened 2021-11-06
Add .tar.zst in .gitignore

RSKYS:
[For later purposes as it can be taken.](https://github.com/SlackBuildsOrg/slackbuilds/pull/1030#issuecomment-962282412)


willysr:
merged

PR #1041, opened 2021-11-10
gis/gdal: Updated for version 3.4.0

ArTourter:
- Shared library .so-version bump.
- Removed explicit disabling of grib driver since the syntax was changed
  in version 3.0 and grib driver was compiled in since.
- Removed explicit path of enabled features since they are detected
  automatically correctly.


willysr:
merged

PR #1052, opened 2021-11-16
Updated anydesk, wimlib, and powershell to their latest versions.

RezaT4795:
network/anydesk: Updated for version 6.1.1.
libraries/wimlib: Updated for version 1.13.4.
system/powershell: Updated for version 7.2.0.


willysr (commented):
Does it need to use root account for operating? Normally we use special user/group specified in https://slackbuilds.org/uid_gid.txt

RezaT4795:
No, it doesn't. I already know about those. That groupadd and usermod commands are what exactly the daemon tells you to execute, if you run it in verbose mode.

Regards,
Reza Talebi
________________________________

willysr:
but does it work if you use nordvn user instead of root?

RezaT4795:
> but does it work if you use nordvn user instead of root?

I don't know. Haven't tested yet.

slackhead (commented):
Usually root only has system groups.  This is a bit unusual.

slackhead:
The 'rm -rf $SOCK' in the rc script in the nordvpnd_start function seems like it should be inside and at the top of the  if/then clause where openvpn is started.

RSKYS (commented):
> The 'rm -rf $SOCK' in the rc script in the nordvpnd_start function seems like it should be inside and at the top of the if/then clause where openvpn is started.

Some few stuffs need to change too including this tho I've tried it before,
I still lose the $SOCK point despite all these changes for some reason..

RezaT4795:
@willysr I pushed back nordvpn for further investigations. In the meanwhile, please merge those other packages.

willysr:
merged

PR #1060, opened 2021-11-24
Bump/tauonmb

RSKYS:
Hi Willy,
As we spoke
I just got time again and worked on it;
As it's dangerous process, I've done everything as perfect as it should be.

You can take the merge or wait for tauonmb 6.8.1, I don't mind much I can wait and submit newer version in this PR.

Regards, Pouria


RSKYS:
Should I mention that you need re-login for switching audio backend?
I mean reboot or atleast logout is required.

willysr:
i will leave this to other admins who are more competent to review this changes

RSKYS:
Sure, thank you

RSKYS:
Sure, thank you
image

I just remind I already used and tested this method in every aspect, and I'm aware of every changes I made, I even know this under /etc/pulse/client.conf
>  ; daemon-binary = /usr/bin/pipewire-pulse

Which ; doesn't work actually, I just added in case someone wanted to hard point at it.

RSKYS:
Yes, this is finally the release officially

> [Phazor] Now uses high resolution output to get the most out of 24bit tracks. (I suggest using Pipewire as your system audio server to help ensure optimal quality)

[As I already mentioned](https://github.com/Taiko2k/TauonMusicBox/releases/tag/v6.8.1), Pipewire's required for this case. I actually even mentioned this change is required and asked for it at LQ for quiet long time now.

RSKYS:
Okay, there are no major differences from my last push;
It was already clean and completed.
I just replaced their version "1.0" instead of standalone "1" so they don't come confusing and match .desktop versions too.

RSKYS:
Okay, there are no major differences from my last push;
It was already clean and completed.
I just replaced their version "1.0" instead of standalone "1" so they don't come confusing and match .desktop versions too.

If any SBo Admin has question before merge, I'm available via email.

alienbob:
The doinst.sh script of your "pulseaudio-pipe" package calls "removepkg pipewire-pulse", and the doinst.sh script of "pipewire-pulse" package calls "removepkg pulseaudio-pipe". 
I am very much against automatic removal of other packages when a package is installed. It is ugly and not transparent to the unsuspecting user.
I would advise to only create ONE SINGLE package and add a configuration script to the package which can be executed by the user and which allows for switching between pulseaudio and pipewire (at least I think that is the purpose of these packages).
You can make it a script with a commandline parameter like "--enable pulseaudio" or "--enable pipewire" for instance.
-- 
Eric Hameleers <alien@slackware.com>

RSKYS:
Hello @alienbob , It's honor I talk to you myself;
Well, I thought it might awkward idea too, have you tried it tho? Are they working?
> You can make it a script with a commandline parameter like "--enable pulseaudio" or "--enable pipewire" for instance.

That's actually might not be bad idea, so "pipewire-pulse" package with 2 commandline parameters,
Nice, I try to write it, if couldn't make it until few hours later, I gotta work and that'd need another day's spare time
Thank you.

RSKYS:
Hello @alienbob , It's honor I'm talking to you myself;
Well, I thought it might awkward idea too, have you tried it tho? Are they working?
The reason I did this was for user to know what mode they're on.
So the left package would be the current mode.
> You can make it a script with a commandline parameter like "--enable pulseaudio" or "--enable pipewire" for instance.

That's actually might not be bad idea, so "pipewire-pulse" package with 2 commandline parameters,
Nice, I try to write it, if couldn't make it until few hours later, I gotta work and that'd need another day's spare time,
thanks.

RSKYS:
> The reason I did this was for user to know what mode they're on.
So the left package would be the current mode.

Are you also against this part?

RSKYS:
Okay, I go for this with singe --enable and --disable, which --enable is the default mode.

Regards, Pouria

RSKYS:
Okay, @alienbob @willysr I request another review please.

PR #1063, opened 2021-11-24
audio/drumstick update to latest version

clavisound:
Requirement for audio/vmpk. Disabled doxygen. Wants doxygen-1.5 we have 1.9.2 but it fails to find it.


willysr:
merged

PR #1067, opened 2021-11-26
Bump/tauonmb

RSKYS:
Hi,
Polished a bit;
[That PR](https://github.com/SlackBuildsOrg/slackbuilds/pull/1060) became messy.

Also there's no problem with my source link, codeload has problem today.
I once again ask admins to review this.

Regards, Pouria.


RSKYS:
Alright, this is working as expected:

#

image
image

RSKYS:
I've set the variable to $PULSEAUDIO
So instead of "PIPEWIRE=disable" in build option, you need to use "PULSEAUDIO=enable".

alienbob:
> I've set the variable to $PULSEAUDIO So instead of "PIPEWIRE=disable" in build option, you need to use "PULSEAUDIO=enable".

You did not understand my previous suggestion apparently.
I do not like the fact that your script produces two variants of a package with the same name but with exactly opposite results. This makes it impossible to guess what the package will do when it gets installed.

What I would like to see instead is a SlackBuild script which builds exactly one package. Inside that package, you should add a script which can be run by the user. The script should accept a commandline parameter for example: "--enable-pulseaudio" or "--enable-pipewire".
The goal is that the user can install your package and then won't have to rebuild it again. The added script must allow the user to switch back and forth between pulseaudio and pipewire easily, simply by using the correct commandline parameter.

RSKYS:
I'd create program for it if I want it that far advanced, I guess I stop releasing newer versions for now.

RSKYS:
It's not like pkgtool > upgradepkg that I put them all in one directory,
Things will get really messy.

PR #1070, opened 2021-11-28
Bump/tauonmb

RSKYS:
Hi,
I don't know, I didn't honesty plan to [release it this fast;](https://github.com/SlackBuildsOrg/slackbuilds/pull/1067#issuecomment-980786860)
[Source](https://github.com/RSKYS/pipewire-pulse)'s designed by me

### I deserve merge this time!


alienbob:
Your README speaks of "$ pipewire-pulse -e    # --enable-pipewire", similar for theother command you describe.
This needs two changes. First, the installed program is now called "pipe-pulse" and second, you use a "$" prompt which suggests that the user can execute the script and that it does not have global consequences. However when I look at that script, it tells me that it needs to run as root, and root's prompt is "#".
Is there no way that you can make it so that the user and not root can execute this and the changes only affect the user, instead of the whole system? Note that you may not be the only user on your computer.

Also, your README instructs people to add "daemon" to the slackpkg blacklist. I do not see why that should be a requirement.

RSKYS:
> This needs two changes. First, the installed program is now called "pipe-pulse" and second, you use a "$" prompt which suggests that the user can execute the script and that it does not have global consequences. However when I look at that script, it tells me that it needs to run as root, and root's prompt is "#".

Oh yeah that needs to become #

> Is there no way that you can make it so that the user and not root can execute this and the changes only affect the user, instead of the whole system? Note that you may not be the only user on your computer.

Also I don't think it'd be possible, this is system change and if any user can change it, they might only get stuck at Plasma login from sddm in case of KDE users..
I want them to know the risk, just like how you upgrade by slackpkg.

> Also, your README instructs people to add "daemon" to the slackpkg blacklist. I do not see why that should be a requirement.

Ah that,
Yeah daemon is indeed unnecessary but pulseaudio needs to remain.

This probably needs another day, thank you @alienbob

RSKYS:
@willysr Please suggest me SBo Admin to email them for this system test;
I need review requests and I'm not gonna send more that one email per day at very last.

Regards, Pouria

RSKYS:
Hi Willy,
These work flawlessly and I've submitted every necessary changes suggested.
Like previous one, there's also [ready binary to test](https://github.com/RSKYS/pipewire-pulse/releases/download/1.1/pipe-pulse-1.1-x86_64-1red.txz) under my releases.
Please merge.

Regards, Pouria.

RSKYS:
Updated.
[New test binary](https://github.com/RSKYS/pipewire-pulse/releases/download/1.2/pipe-pulse-1.2-x86_64-1red.txz)

pprkut:
Maybe I'm missing something, but why is this installed alongside pulseaudio and not instead of? As far as I can see the `pipe-pulse` script as a whole looks completely unnecessary. All the config files can simply be put in place by the SlackBuild itself and then users should just

```
removepkg pulseaudio
installpkg pipewire-pulse
```

Switching back would be the reverse.

The only reason to install them both alongside each other would be to allow users to choose. But you already mentioned above you don't want to do that...

RSKYS:
@pprkut We're not exactly using distribution with systemd or dependency resolver;
Things are a bit different up here.

I'm not sure if it's possible to write whole new package that takeover PulseAudio functionality without daemon easily,
and that way I only have to re-package whole PulseAudio with different name and different configs
Also I want to be able to add it as dependency of "tauonmb", so if users don't like it and for example using sboui, they wouldn't have to annoy themselves with adding removing dependencies.
I already thought about that.

pprkut:
Alright, that's fair, but then the slack-desc is misleading.

```
Replacemnet of PulseAudio backend
```

Is not true. `pipe-pulse` isn't a replacement for pulseaudio. It's a tool to switch from pulseaudio to pipewire.

RSKYS:
> Is not true. pipe-pulse isn't a replacement for pulseaudio. It's a tool to switch from pulseaudio to pipewire.

Oh that's right,
well don't blame me, I've re-written this package for three times;
It changed a lot from first time.

RSKYS:
@pprkut It's fixed now
So if it doesn't have any other problem, can it get merged?
Thanks

PR #1106, opened 2022-01-03
Merge request from "red/master"

RSKYS:
Late merry christmas I guess!
These have released for quiet some while.
#
Updates:
- Kvantum 1.0.0 now can get compiled by qt6, but it continues to compile by qt5 since there's no need for me to worry about new Qt pushes so soon, and I don't have to change it's name, **yet.**
- Nix 2.5.1
- Picard 2.7.1
- Natsort 8.0.2

I haven't decided how or when should I update Tauon;
I've even pushed pre-release to my own repo, but I don't know what should I do about official,
Not that I don't know,
***Admin(s) didn't really leave me with any good memory of it.***


willysr:
merged

willysr:
@RSKYS we just found out some issues about NIX
- it install nix and some libraries in filesystem during build, which shouldn't happened
- man pages is not properly compressed
- main nix binary is not stripped
- doesn't build on 32 bit architecture

PR #1111, opened 2022-01-06
network/teamviewer: Readded and updated for version 15.25.5

ArTourter:
It looks like teamviewer requires it license files to be located in
/opt/teamviewer/doc so moving these files to the standard slackware path
(/usr/doc/$PROGNAM-$VERSION) prevents it from working.

Here I basically do not move these files but we could also create a
symlink between the slack path and the teamviewer path. Not sure which
would be prefered.

Also I am not particularly asking to take over the maintenance of this.
just that I managed to fix it on all my machines and thought I would
submit it back for testing and discussion.


ArTourter:
btw 15.26.4 also works fine.
I was also wondering if it is necessary to keep the RTlibs folder. in contains libraries already available in slackware. I have tried removing the folder and every thing still works fine. Doesn't save much though.

willysr:
Did you test this on runlevel 3?
i never able to get the GUI up and running if i use runlevel 3

ArTourter:
Willy, no I hadn't tested it on RL3. I have just tried it (with and without the RTlib folder) and indeed, it doesn't launch the GUI.... hmm.
there is an interesting post on LQ about the differences between the 2 but I really don't know what to do with it.
I guess we could put something in the README for the time being about requiring RL4... but it is not really satisfying

willysr:
I have one people confirmed that it's working on RL4, but since the repo is out, this can't be added directly via git and must be submitted via submission form. I will keep this PR open until someone send a submission.

ArTourter:
@willysr are you happy for me to take over since you were originally the maintainer?

willysr:
sure :)

ArTourter:
Thanks @willysr . I have submitted my script via the web form as requested. It is for the latest version 15.28.8.

PR #1143, opened 2022-02-04
For s bo

duncan-roe:
WillySR maintains libraries/gf2x but I had a need to build it recently and got *403 Forbidden* from the download link.
Googled for new home and updated gf2x.info
Patch gf2x.SlackBuild:
 - Rename untarred directory back to how it used to be
 - Need autoreconf with this packaging
 - Bump BUILD since software version is unchanged


willysr:
mkchroot patch does not apply, probably it has been updated and you haven't rebased to latest changes in master?

willysr:
i took the gfx2 patch already, so just waiting for mkchroot

PR #1148, opened 2022-02-04
libraries/libxdg-basedir: Update for version 1.2.2.

newHeiko:
Also switched to a better maintained fork.

This version already includes the patch.


willysr:
merged

PR #1150, opened 2022-02-05
audio/quodlibet: Updated for version 4.4.0.

nick8325:
The version of quodlibet currently on SBo doesn't work on Slackware 15, but this new version does.


willysr:
merged

PR #1160, opened 2022-02-06
academic/sage-bin: Updated for version 9.4

duncan-roe:
This will be the last sage-bin update because upstream have discontinued binary packaging.
The 9.4 package had symlinks to system gcc files with a non-Slackware path name so these are supplied in debian_binaries-11.2.tar_xz (from a debian VM I created).
Also, when sage does its initial patching run,  9.4 needs _libbrotli{enc,dec,common}_ (with _common_ having to be in a separate directory, not kept by the SB).
The patches avoid warning about deprecated escapes. I verified that at least some of the patches are in 9.5 (from git).


willysr:
Hi @duncan-roe is it possible to put the tar_xz files outside of the repository and put it in the .info files?

duncan-roe:
Hi Willy,

On Sat, Feb 05, 2022 at 10:03:11PM -0800, Willy Sudiarto Raharjo wrote:
> Hi @duncan-roe is it possible to put the tar_xz files outside of the repository and put it in the .info files?

debian_binaries-11.2.tar_xz is <2K so not worth worrying about I'd suggest.

debian_libraries-11.2.tar_xz at 270K would add 25% to the repo size so
can we put it in the repo you have for big files? Then I could list it
on the DOWNLOAD line.
Cheers ... Duncan.

willysr:
you want to put it on SF? i can add you to the developer role so you can add your files there if you like
just send me your username via private channel :)

PR #1166, opened 2022-02-07
Red/master

RSKYS:


willysr:
python3-protobuf is same as protobuf3 so it's duplicate
by the way, newly scripts should wait until submission is open
we are busy with fixing the scripts to make sure it's working with 15.0

RSKYS:
Hey,
It's working, I'll test them before I commit here on my own repo.
And yeah, protobuf3 is yours it seems :)
So, are other 3 okay?
Cause zeroconf is dependency of pychromecast and pychromecast is new dependency of Tauon;
If other 3 are okay with you, Imma fix em right away.

RSKYS:
Okay, I have doubt and question..
the main part I need in protobuf is the [python part](https://github.com/protocolbuffers/protobuf/tree/master/python),
Does that mean I need full "protobuf3" to reach the requirement? Cause pip3 can satisfy without it, and I think python3-protobuf is enough.

willysr:
we don't test using pip3, because it could mess up with the rest of the installation and it's not standard Slackware way of handling packackes

RSKYS:
Huh,
It's not my standard way neither,
that's why I package them..
Well, I don't know what should I do with it for now, I exactly committed these to my repo and only thing that matters is it's working;
I'll decide what should I do with official later.

PR #1170, opened 2022-02-08
ghostwriter update 2.1.1

afhpayne:
Taking over with permission from Andrzej Telszewski


willysr:
merged with some modifications

PR #1187, opened 2022-02-09
academic/sage-bin: Updated for version 9.4

duncan-roe:
This will be the last sage-bin update because upstream have discontinued binary packaging.
The 9.4 package had symlinks to system gcc files with a non-Slackware path name so these are supplied in debian_binaries-11.2.tar_xz (from a debian VM I created).
Also, when sage does its initial patching run,  9.4 needs _libbrotli{enc,dec,common}_ (with _common_ having to be in a separate directory, not kept by the SB).
The patches avoid warning about deprecated escapes. I verified that at least some of the patches are in 9.5 (from git).
*** I left the tar.gz files on SF under source control because I couldn't figure out how to upload them otherwise


willysr:
Merged and i fixed the source files in SF

PR #1188, opened 2022-02-10
python/python-dbus-next: Upload python-dbus-next

isaackwy:
python-dbus-next is an optional dependency for qtile - it sends notifications through dbus for qtile.

I updated the qtile README accordingly as well.
A notification daemon (e.x. dunst) is required for showing notifications on qtile.


willysr:
merged

PR #1191, opened 2022-02-11
Upload python-pep517, python-build and python-install

isaackwy:
python-build and python-install are for building Python packages that have a pyproject.toml file and do not have a setup.py file.
python-pep517 is a dependency of python-build.

For example, to build a package, place the following lines in a .SlackBuild file:
python3 -m build --wheel --skip-dependency-check --no-isolation
python3 -m install --optimize=1 --destdir=$PKG dist/*.whl


willysr:
this is not dependency for existing scripts right?
if that's the case, i will postpone it until we have 15.0 ready

isaackwy:
They're not build dependencies for existing scripts, but rather build dependencies for later versions of packages.

I asked for permission to take over maintenance of python3-jsonschema from Benjamin Trigona-Harany <slackbuilds@jaxartes.net>.
I plan to update it from 3.2.0 to 4.4.0.
The source tarball only has a pyproject.toml file, but does not have a setup.py file. I will need python-build and python-install to build python3-jsonschema 4.4.0.
I would prefer python-pep517, python-build and python-install to be uploaded, but please postpone uploading the 3 packages if you would like.

willysr:
Is it not enough with this package that has been added to Slackware?
l/python-tomli-1.2.2-x86_64-1.txz: Added.
       This is needed by setuptools_scm. Also, the SlackBuild contains a nice
       example of using a setup.py shim to setuptools for a Python module that
       only contains pyproject.toml and not setup.py.
       Thanks to Heinz Wiesinger.

isaackwy:
You mean something like the following?

> cat << EOF > setup.py
from setuptools import setup
setup(name="${SRCNAM}", version="${VERSION}", packages=["${SRCNAM}"], package_data={"": ["*"]})
EOF
unshare -n python3 setup.py install --root=$PKG || exit 1

Thank you for the advise Willy. Please close this pull request without uploading python-pep517, python-build and python-install then.

PR #1193, opened 2022-02-12
Update skalibs, execline, s6, s6-rc, s6-linux-init, and mailspring

mumahendras3:
Mostly version bump changes except for s6 which includes moving root-only binaries to /sbin.


willysr:
merged

PR #1196, opened 2022-02-12
libraries/openjfx: Remove.

newHeiko:
As mentioned on slackbuilds-users:

Not required for multimedia/MediathekView any more.

No updates on the binary package I've been repackaging for years.

Can be easily replaced by an Azul build of OpenJDK with JavaFX, if
anyone needs JavaFX.


willysr:
merged

PR #1200, opened 2022-02-14
Update libtrash and sfeed. Upload pass-import

isaackwy:
I am maintaining all of these packages below.
Please update libtrash and sfeed, and please upload pass-import.

Edit: I revised the scripts - I switched /bin/sh back to /bin/bash


willysr:
merged

PR #1202, opened 2022-02-15
perl/perl-Locale-Codes: new package

mab974:
Signed-off-by: Michel Begue <mab974@gmail.com>


willysr:
is this a new dependency?
if not, i would suggest to wait until 15.0 is ready and submit it via normal submisison form

willysr:
is this a new dependency?
if not, i would suggest to wait until 15.0 is ready and submit it via normal submisison form

mab974:
yes, due to perl version change

PR #1204, opened 2022-02-15
perl/perl-Gtk3-SimpleList: new package

mab974:
Signed-off-by: Michel Begue <mab974@gmail.com>


mab974:
old one which is not in the list !?
can wait

PR #1206, opened 2022-02-16
gis/gdal: fix building when freexl is installed.

ArTourter:
Don't use libtool for building the libraries. Otherwise the build
process looks for freexl .la file which is no longer packaged.


willysr:
merged

PR #1211, opened 2022-02-16
python/python3-openpyxl: Fix MD5SUM

isaackwy:
Fellype (fellype@gmail.com) just found that the md5sum in the .info file does not match the source md5sum.
I have sent in a fix to the .info file.


willysr:
merged

PR #1212, opened 2022-02-17
shapely update

ArTourter:
Shapely was renamed shapely and moved to a new github project.
I changed the name of the package and updated the info file for all the dependees as well as tested them.
I know that 2 of these (i.e geopandas and gj2ascii) have not yet been picked up by a new maintainer following Benjamin dropping them but since they are still there I updated them anyway.


ArTourter:
Yup, and done!

ArTourter:
Hmm looks like someone is changing things while I am working on it.
I wish people would contact the maintainers when they are very active before changing things.

ArTourter:
BTW there is inconsistency about this since it is still called Shapely on pypi but not on github.
Since the gihub files have been regenerated and therefore have a different hash, but the ones on pypi haven't I suggest we do not change the package name yet and use the source from pypi:

https://files.pythonhosted.org/packages/source/S/Shapely/Shapely-1.8.0.tar.gz f2b43b00b9eccec76af24a40201a8223

then we don't have to change any other packages yet.

willysr:
please talk to Andrew since he was the one who made the changes

aclemons:
Apologies @ArTourter, didn't mean to step on your toes. We have list of over 600 packages still failing to build for various reasons. I did see you were active from the git history, but simply fixed this build while going through the list since the MD5SUM had changed when the project was renamed. I've reverted this out of my branch and willysr will merge your other PR (https://github.com/SlackBuildsOrg/slackbuilds/pull/1230) which changes to the pypi URL instead. I am subscribed the mailing list, but somehow I had overlooked your email. Again, apologies - we are all just volunteers here trying to get the repo out for 15.0.

ArTourter:
@aclemons Sorry I didn't mean to come at you that sharply. Thanks for your work!

I am still wondering whether to change the name and even maybe prefixing it with python3- like other python packages. but I am still puzzled about the inconsistency in naming.

PR #1225, opened 2022-02-19
Upload ipython (+new maintainer)

isaackwy:
I will maintain ipython in Benjamin Trigona-Harany's place \<slackbuilds@jaxartes.net\>.
I will also take over maintenance of ipython_genutils, python3-jedi and traitlets (dependencies of ipython).
I have uploaded python3-matplotlib-inline as an additional dependency of ipython.

ipython's latest version is at 8.0.1; however, it requires many more dependencies (e.x. python-black).
Therefore, I have kept ipython at 7.31.1.
In addition, for traitlets, I removed ipython_genutils from the list of dependencies (I have updated traitlets to 5.1.1, and this new version does not require ipython_genutils).


willysr:
merged

PR #1226, opened 2022-02-19
network/kismet: Updated for version 2022-02-R1 + new maintainer.

pyllyukko:
Maintainer change agreed with @audriusk.


willysr:
please make the VERSION using _ instead of - since it will messed up with slackpkg
so it should be 2022_02_R1

pyllyukko:
I'm pretty sure the `${VERSION//-/_}` should take care of that ([here](https://github.com/pyllyukko/slackbuilds/blob/968fb30257603f91d0672bb87685586674e683cb/network/kismet/kismet.SlackBuild#L132)). So the end result will be `kismet-2022_02_R1-x86_64-1_SBo.tgz`.

willysr:
But then the VERSION will not match the resulting package :)

pyllyukko:
Are you talking about `sbopkg` (instead of `slackpkg`)?

pyllyukko:
I made it into `2022.02.R1`, so it will also look nice in [Repology](https://repology.org/project/kismet/versions).

pyllyukko:
I made it into `2022.02.R1`, so it will also look nice in [Repology](https://repology.org/project/kismet/versions).

pyllyukko:
Fixed the `rm -rf $PRGNAM-$VERSION` also into `$PRGNAM-${VERSION//./-}`.

willysr:
merged

PR #1242, opened 2022-02-23
libraries/GtkD: Removed (not used by any slackbuild)

philipvdh:
This was only used by tilix which has been removed also.


willysr:
merged

PR #1243, opened 2022-02-23
Upload jupyter-qtconsole and update QDarkStyle

isaackwy:
jupyter-qtconsole (and its dependency QtPy) are SlackBuilds previously maintained by Benjamin Trigona-Harany \<slackbuilds@jaxartes.net\>.
I would like to take over maintenance of QDarkStyle as well, with permission from Dimitris Zlatanidis \<d.zlatanidis@gmail.com\>.
I have uploaded python-helpdev as a new dependency of QDarkStyle.


willysr:
merged

PR #1246, opened 2022-02-24
games/lgogdownloader: updated for version 3.9

Ythogtha:
I also replaced the replacement URL with something probably more suitable, since development seems to happen on github.
And reinstated the warning in README as it still cannot build without a static rhash (I've just tried) whatever the reason.


willysr:
merged

PR #1247, opened 2022-02-24
games/vcmi-resolutions: updated for version 20151113

Ythogtha:
This actually only reflects the real date of the file downloaded, that change should have happened more than 6 years ago.
As for all vcmi-{addons}, it only downloads and repackage the file on http://download.vcmi.eu/ and the only pertinent version is the date the file was added there.


willysr:
merged

PR #1248, opened 2022-02-24
games/vcmi: updated for version 0.99_git20220130

Ythogtha:
I also changed the version pattern : better to use git{date} in the version than short hash, since hashes can't be properly compared to one another. The exact commit is still perfectly visible on the download URL and vcmi.Slackbuild

I might separate between vcmi@0.99 and vcmi-dev@git-revision, but I'm not really sure there is a point there.
Since I can't play properly (various crashes, using proper and uptodate GoG Homm3 datas) with any of those builds, I guess it's just instability on upstream project, all I am prepared to do for now is try and give the most recent build, hoping for the best.


willysr:
merged

PR #1249, opened 2022-02-24
academic/gpredict: revert to latest stable version 2.2.1

Ythogtha:
2.3 is a development release, the file itself is prone to change on github, it is not proper for a slackbuild.
2.2.1 is the actual latest release, and furthermore, it is more recent than the 2.3 snapshot.


willysr:
merged

PR #1250, opened 2022-02-24
psycopg2 updated for 2.9.3, legacy python2-psycopg2 added

Ythogtha:
So that one is a test for workflow, I will have others (quite a few I think, and a lot of work too) following the same pattern, so I crash-test the workflow on that one to make things go smoothly for the following ones.

Problem is:
* psycopg2 is stuck at current version 2.8.6 for python2, and current Slackbuild builds for both python2 and python3.
* I want/need to update it for python3.
* there's still one package requiring it at python2, hence I can't simply drop python2 support there.
Hence there are three commits:
* I created a new python2-psycopg2, copying stuff from the actual psycopg2, which creates only python2 module.
* I updated business/trytond REQUIRES for python2-psycopg2.
* I updated psycopg2 for 2.9.3, and removed creation of python2 module.

The repository should be fully functional after that, but the tags on the website would not be added for python2-psycopg2.


willysr:
merged

willysr:
These scripts depends on the original psycopg
gis/geopandas
gis/qgis
python/odoo
system/barman

did you test them all?

Ythogtha:
Yes, I did, I managed to build them all, and start the programs or load the python modules.
odoo complains about missing werkzeug.contrib, so I can't know if psycopg2 will cause problem later on.

PR #1260, opened 2022-02-24
Update python-prometheus_client

isaackwy:
I am now the maintainer of python-prometheus_client. I have updated the python-prometheus_client SlackBuild to 0.13.1.
EDIT: python-prometheus_client >=0.13.0 drops Python 2 support. Therefore, my SlackBuild also drops Python 2 support.


willysr:
merged

PR #1269, opened 2022-02-27
Afdko

duncan-roe:
afdko build requires scikit-build which is a new SB. Keywords python scikit-build skbuild. Important to include skbuild as a keyword because that's what gets reported as missing but PyPi search doesn't find it - I had to google to get package name.
afdko build is substantially changed to use CMake.


willysr:
hey @duncan-roe python3 is now included in Slackware 15.0, so it should be enabled by default, no need to run check. Also please use python2 just for consistency if you want to build python2 modules
slack-desc also need to be fixed as it didn't comply with our template
this is my comment on scikit-build

PR #1270, opened 2022-02-27
15.0 updates - 6 packages

bassmadrigal:
Hopefully you're fine with a lump update. Most are version bumps except for me reverting ownership of multimedia/picard. All tested against 15.0.


RSKYS (commented):
I have no problem with ownership revert of "picard".

It needed some major changes, like the fact it wasn't working with python2 anymore, neither dependencies;
I won't stop bumps and fixes tho if it would be necessary!

willysr:
merged

bassmadrigal:
> I have no problem with ownership revert of "picard".

Good, it wasn't yours too take over in the first place.

> It needed some major changes, like the fact it wasn't working with python2 anymore, neither dependencies; I won't stop bumps and fixes tho if it would be necessary!

You should be reaching out to me, not **_lying_** saying you have and that I'm not using Slackware anymore. We had even talked weeks earlier about discord, so you knew I was active.

You should not be pushing updates for packages you don't maintain without permission from the maintainer.

I've had a working update for some time, but I didn't push it to the 14.2 repo since it required qt5 and there wasn't much change in the actual program's functionality. However, if I knew someone wanted the update, I could've pushed it anyway.

This one really pissed me off by you lying about contacting me and then lying saying that I'm no longer using Slackware. I've been using Slackware for close to twenty years now and I have no intention on switching away from it.

PR #1275, opened 2022-02-28
Buildbot worker

Ythogtha:
That one was a bit more complex at first sight.
buildbot-slave has been renamed buildbot-worker quite some time ago, and slackbuild version is very obsolete.
I don't care about trying it for python2, I'm going straight for python3, because what would be the point ?
That means cloning some python2-only packages for python3.

So : replacing buildbot-slave by buildbot-worker, and creating python-twisted, python-automat and python-zope.interface for python3 versions of those builds.


Ythogtha:
This needs a bit more work...

PR #1277, opened 2022-03-01
Buildbot worker

Ythogtha:
That one was a bit more complex at first sight.
buildbot-slave has been renamed buildbot-worker quite some time ago, and slackbuild version is very obsolete.
I don't care about trying it for python2, I'm going straight for python3, because what would be the point ?
That means cloning some python2-only packages for python3.

So : replacing buildbot-slave by buildbot-worker, and creating python-twisted, and python-automat for python3 versions of those builds.


Ythogtha:
That one is ok, builds fine on a clean 15.0 install, no missing REQUIRES, and proper versions.

willysr:
you are still using the old python-zope instead of python3-zope

Ythogtha:
Damn, sorry, should be better now...

willysr:
the commit message still hasn't been changed

Ythogtha:
should be better now...

PR #1281, opened 2022-03-02
2 Updates 1 Addition

bassmadrigal:
These updates/addition are to get pitivi working on 15.0 with the newer gst version.

libraries/gsound: New dependency for pitivi not already on SBo.
libraries/gst-editing-services: renamed from gstreamer-editing-services to match upstream name and version bumped to match  15.0's gstreamer version.
multimedia/pitivi: version bump to latest version

I imagine both pitivi and gstreamer-editing-services were ftbfs and this should fix them.


bassmadrigal:
Got another version bump done for puddletag and then I realized I kept the PRGNAM and short desc at the top of the README for gsound.

willysr:
As for gsound, can you make it comply with our newest template?
https://git.slackbuilds.org/templates/tree/meson-template.SlackBuild
It should use /bin/bash instead of /bin/sh

You can also remove the unneeded code like the 3 find part

bassmadrigal:
Oops, apparently I need to update my SlackBuild generator script with the 15.0 templates. Update has been pushed.

bassmadrigal:
Got a few more simple version bumps.

In the future, would you prefer different updates in a different PR if you haven't merged the initial one yet? I'd like to keep this as easy as possible for you.

aclemons:
> I imagine both pitivi and gstreamer-editing-services were ftbfs and this should fix them.

Excellent - yes I had pitivi on my list and had seen that basically the only fix was a version bump + new dependencies. Thanks for this.

> In the future, would you prefer different updates in a different PR if you haven't merged the initial one yet? I'd like to keep this as easy as possible for you.

For this PT at least, I think it is sensible to keep it focussed on the pitivi fixes and dependencies

libraries/gst-editing-services/gst-editing-services.SlackBuild
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
```
aclemons, line 1: This one ftb for me:

```
[...]
cp: cannot stat 'INSTALL': No such file or directory
```

aclemons (changes requested):
Could you recheck gst-editing-services?

bassmadrigal:
> Could you recheck gst-editing-services?

Oops, not sure how that slipped through. It's been updated.

aclemons:
All merged on my branch now. Thanks!

PR #1282, opened 2022-03-02
Red/master

RSKYS:
Rebased #1268


willysr:
Does windscribe only works for 64 bit architecture?
how about 32 bit system?

RSKYS:
Hi,
No Willy, there's no support for 32bit version.

willysr:
are you sure? the project page doesn't mention about this but it was tested against older Ubuntu version which still have 32 bit support

RSKYS:
Mhm.. current binary?
I don't have 32bit system to test tbh ¯\\\_(ツ)_/¯
There was only old thread on reddit and I assumed dead, if you can verify this binary works with 32bit, I'll apply fixes for /opt and I'll move source to simple DOWNLOAD in info file.

RSKYS:
Either way, this might be better for now.
Let's investigate 32bit another time.

willysr:
merged
please don't change the SlackBuild permission to 755

RSKYS:
Oops, I just saw where was wrong,
sorry xD

PR #1283, opened 2022-03-02
python/python-automat: cloned from Automat for python3, updated for version 20.2.0

Ythogtha:
Requirement of python-twisted (python3)


willysr:
if it's for python3 specifically, please use python3-automat as the package name

Ythogtha:
I was following Slackware's naming pattern, since nothing has been clearly decided on the slackbuild.org mailing list, and I participated on those discussions.
Which is python-package for current python packages, and python2-package for obsolete python2 packages.
And sometimes just package for non-module packages (like mercurial, or in my case buildbot, buildbot-worker, and later on Kallithea).

willysr:
we did rename most of the packages to use the convention
python2-xx for specific python2 modules
python3-xx for specific python3 modules
python-xx for python2 AND 3 modules

willysr:
Slackware only backport some core and important python2 modules and put them into a single package and the rest will go with python3-xx, but here in SBo there are scripts that still need python2, so to make it easier for future python2 removal, it would be easier if we can start splitting it now rather than later

Ythogtha:
I totally agree with splitting sooner rather than later, and naming python2 packages python2-xx!
And trying to remove them as soon as they are no more useful (python2-twisted might take some time here...)

I just disagree on current python package naming. Current python is python3, so python-xx should build for python3, there are no packages named python3-xx in Slackware (except python3 itself).

But if that means finding more easily which packages aren't splitted, and hunt down all python-xx packages that builds for both python versions, then there is a point in naming python3-xx for already splitted ones.

willysr:
there will be time where python 4 is going to be released :)

Ythogtha:
Yes, let's hope we'll have learned and make less of a mess, and be more prepared, that time :)
Ok, renaming in place with force-pushes, to python3-xx.
I'll have to push updates to buildbot and buildbot-worker so they REQUIRE python3-xx dependencies.

Ythogtha:
All done, 6 PR ready for python3 !

willysr:
merged

Ythogtha:
Thanks a lot for your time Willy !

PR #1300, opened 2022-03-04
Upgrade bumblebee-status to last version and remove patch now useless.

Tonus1:
As stated in title.
It's my first pull request ever so tried on a harmless object.


willysr:
merged

PR #1302, opened 2022-03-04
office/xournal: tested; builds on 15.0

brobr:
Hi, saw that xournal was on the list of non-builders but what is the problem? Builds and works fine at my end..


willysr:
confirmed

PR #1303, opened 2022-03-04
libraries/htslib: Update to version 1.15

brobr:
Update to version 1.15


willysr:
merged 
please don't remove the .la removal line since Slackware doesn't ship .la files anymore, so we are going to remove it as well

brobr:
On Fri, 04 Mar 2022 17:17:58 -0800
Willy Sudiarto Raharjo ***@***.***> wrote:

> merged 
> please don't remove the .la removal line since Slackware doesn't ship
> .la files anymore, so we are going to remove it as well
> 
> 

Sorry, was not aware of this; has this been put in the templates
already?

Thanks for all the work; please keep it up.

Rob

PR #1306, opened 2022-03-04
academic/pysam: Use internal htslib-1.14

brobr:
After htslib update to version 1.15 pysam needs to use the internal htslib 1.14


willysr:
merged

PR #1308, opened 2022-03-05
gis/python3-xarray: renamed from python/xarray and updated for version 2022.03.0

ArTourter:
This is very much a gis package and is python3 only so rename now before
other packages start depending on it.


willysr:
merged

PR #1310, opened 2022-03-05
system/tbsm: Update for 0.6

isaackwy:
I also made a few changes to README.SLACKWARE as well.
Loading consolekit is no longer necessary for desktop environments - I removed that mention from README.SLACKWARE.
I also changed the example window manager mentioned in README.SLACKWARE from i3 to dwm.


willysr:
merged

PR #1315, opened 2022-03-06
python/ipython: Update for 8.1.1 (+new stack_data dependency)

isaackwy:
I would like to update ipython for 8.1.1.
stack_data, pure_eval and python-executing are new dependencies for ipython >= 8


willysr:
merged

PR #1318, opened 2022-03-07
system/Type1_to_OTF: Updated for Slackware 15.0

duncan-roe:


Ponce:
Hi @duncan-roe , it's not allowed to install packages in the system from a SlackBuild.
I understand you do this because on SBo there aren't the dependencies you need but you should actually submit them too...

willysr:
It's probably best to remove this script for now and wait until you have all the deps submitted before you resubmit this script again

Ponce:
> It's probably best to remove this script for now and wait until you have all the deps submitted before you resubmit this script again

👍

PR #1331, opened 2022-03-08
Kallithea

Ythogtha:
This is the goal of my last big work for 15.0 : Kallithea and its gazillions python3 dependencies.
I'll rename some packages, and remove python2 wherever I can (ie: when nobody else needs it).


Ythogtha:
And that's 20 commits, still not finished but I'm calling it a day.
Sorry for all the work, but I'm almost over, yay :)

willysr:
merged

PR #1345, opened 2022-03-08
Python3 dulwich

Ythogtha:
Separating older dulwich to allow for updated new python3-dulwich.
Updates on hg-git which requires python3 dulwich, and doesn't work worst with this new version (whatever: it requires an update for slackware 15.0).
Updates on older dulwich which was python2/python3, to stay python2 only.


willysr:
merged

PR #1348, opened 2022-03-08
network/sfeed: Update for 1.3

isaackwy:
I changed the homepage of the SlackBuild as well:
https://codemadness.org/sfeed-simple-feed-parser.html


willysr:
merged

PR #1350, opened 2022-03-08
python/python3-paginate: added, requirement for Kallithea

Ythogtha:


willysr:
merged with some modifications

PR #1358, opened 2022-03-08
python/python3-urlobject: renamed from URLObject for python3

Ythogtha:
Flask-Dance: Updated REQUIRES for python3-urlobject as it is python3 only


willysr:
merged

PR #1367, opened 2022-03-09
network/et: added cmake option -DDISABLE_VCPKG=ON

khronosschoty:
Apparently et was downloading during build, adding -DDISABLE_VCPKG=ON disables et build system from downloading dependencies.


willysr:
merged

PR #1368, opened 2022-03-09
python slackbuilds renames and updates

ArTourter:
Thought I'd do some cleaning before the official 15.0 release so:
appended python3- to my python slackbuilds and updated the dependencies in the dependees.


willysr:
merged

PR #1369, opened 2022-03-09
cargo: avoid writing to $HOME/.cargo

pghvlaans:
Unless `CARGO_HOME` is set explicitly, cargo will create `$HOME/.cargo/.package_cache` for the running user if it does not already exist.

There may be more scripts that cause this behavior, but these were the ones I was able to find.


willysr:
merged

PR #1375, opened 2022-03-13
network/netcat-openbsd: Update download URL

duncan-roe:
I had an email that the old URL had stopped working with the message "the given path has multiple possibilities: ...". Found this to be true: new URL suggested in email is good.


willysr:
merged

PR #1376, opened 2022-03-13
Update skalibs, execline, s6, s6-linux-init, and mailspring

mumahendras3:
Bump skalibs, execline, s6, s6-linux-init, and mailspring versions to their latest versions.

Thanks!


willysr:
merged

PR #1377, opened 2022-03-14
development/rstudio-desktop: Update for 2022.02.0+463 (+new maintainer)

isaackwy:
Ekin Akoglu \<ekinakoglu@gmail.com\> provided me permission to take over maintenance.

Newer versions of the RStudio binary do not run - they receive a segmentation fault from /lib64/ld-linux-x86_64.so.2
Therefore, I have built RStudio from source.

I have uploaded soci (a new dependency of RStudio) as well.

In addition, newer versions of RStudio run only on 64-bit systems.
I have uploaded a rstudio-desktop-legacy SlackBuild, which supports 32-bit systems. rstudio-desktop-legacy is at version 1.1.463 (i.e. the version currently available at SlackBuilds.org)


willysr:
new dependency should go to normal submission form, so i can't approve this PR

PR #1385, opened 2022-03-14
python/anyjson: Removed, very old and not in any more REQUIRES

Ythogtha:
Old REQUIRE of other stuff, not needed anymore anywhere, not useful, and not updated upstream since 2012.


willysr:
merged

PR #1386, opened 2022-03-14
python/WebHelpers: Removed, replaced by WebHelpers2 on python3-webhelpers2

Ythogtha:
Slackbuilds wich used to REQUIRE this upgraded to WebHelpers2, hence that one is unneeded, no need for it to stick around.


willysr:
merged

PR #1394, opened 2022-03-15
python/python3-service-identity: cloned from service_identity for python3, updated for version 21.1.0

Ythogtha:
Added, requirement for python3-twisted


Ythogtha:
I'll do that as soon as it is reopened.

PR #1399, opened 2022-03-15
Hugo update

afhpayne:


willysr:
hugo is at 0.92.2 on master, so this patch doesn't apply cleanly

PR #1402, opened 2022-03-16
Update jupyter-notebook for 6.4.10 (+update deps)

isaackwy:
1. The git pull request for updating jupyter-nbconvert to 6.4.3 was already merged.
However, I would like jupyter-nbconvert to be updated to 6.4.4.
EDIT: jupyter-nbconvert is now based on master.

2. In addition to updating jupyter-notebook to 6.4.10, I have removed the mathjax2 dependency from the .info file.
Jupyter Notebook already has mathjax2 bundled.


willysr:
jupyter-nbconvert is not based on master, so the patch failed

willysr:
merged

PR #1405, opened 2022-03-18
nextcloud-desktop update

afhpayne:


willysr:
nextcloud-desktop is new script, so it should be submitted via submission form (which is still closed), but i will take brave-browser update

willysr:
ohh turns out brave is already updated before

PR #1408, opened 2022-03-19
python/python2-twisted: reverted to version 17.9.0

Ythogtha:
I made a mistake on this update, to upgrade version, I need new packages...
And all that is pretty untested.
Sorry about that !


willysr:
merged

PR #1411, opened 2022-03-21
Rclone 1.58.0

afhpayne:


willysr:
can you remove the previous 3 commits? it should contain only atomic commit

PR #1418, opened 2022-03-24
Please update python-evdev, qtile and ipython

isaackwy:
python-evdev is now at 1.5.0
qtile is now at 0.21.0
ipython is at 8.2.0


willysr:
merged
don't change the permission

PR #1435, opened 2022-03-30
perl/MoarVM: Updated for version 2022.03

mab974:
email address also changed.


willysr:
libuv is now part of Slackware 15.0
Please make sure that you test it on clean Slackware 15.0 full installation

PR #1437, opened 2022-03-30
perl/rakudo: Updated for version 2022.03

mab974:
email address also changed.
MoarVM, nqp and rakudo works together.


willysr:
merged

PR #1442, opened 2022-03-31
graphics/gscan2pdf: Updated for version 2.12.4

mab974:


willysr:
merged
don't put optional deps on slack-desc
put them on README

PR #1443, opened 2022-03-31
python/python-m2r:  Removed: Unused, obsolete, break if msitune is upgraded

Ythogtha:
I think python-m2r was historically a requirement of Automat, when I was not maintainer.
It turns out it is not required at all.
Besides, it is obsolete, unmaintained, and prevent upgrading of mistune.
Hence I removed it from python2-automat and python3-automat, and removed it from Sbo altogether.


Ythogtha:
My mistake : I'm not the owner of python-m2r, I'll contact him.

PR #1448, opened 2022-04-02
gis/gdal: fix download file.

ArTourter:
using the release file from github rather than the archive source.
Somehow the archive source changed since my submission, and the archive
source has a broken symlink.
Can this be pushed quickly as currently the slackbuild doesn't work.


willysr:
Fixed in web and git, but tarball and SlackBuild script will be updated next public update

PR #1449, opened 2022-04-02
desktop/xss-lock: fix cmake policy obsoloscence warning

pghvlaans:
Nothing major for now, but the warning might turn into actual trouble at some point as -current upgrades cmake.


willysr:
merged

PR #1450, opened 2022-04-02
Update piper and jupyter-ipykernel (HOMEPAGE and README)

isaackwy:
system/piper: HOMEPAGE should be https://github.com/libratbag/piper
development/jupyter-ipykernel: version 6.10.0 is the last possible version


willysr:
merged

PR #1453, opened 2022-04-03
python/python2-twisted: updated for version 20.3.0

Ythogtha:
This update needs the new SlackBuild for python2-hamcrest to be added.
It has been sent through proper submission channels.


willysr:
merged

PR #1456, opened 2022-04-04
nextcloud-desktop update 3.4.4

afhpayne:


willysr:
@afhpayne please don't change the changes we made regarding slack-desc, makepkg command, and URL in .info. That format will give you consistent naming, so you don't have to use tar xvf $CWD/v$VERSION.tar.gz || tar xvf $CWD/$SRCNAM-$VERSION.tar.gz

PR #1457, opened 2022-04-04
Mab974

mab974:
Hi Willy,

one slackbuild per commit this time.
tell me if it’s okay. 
I kept the . patch files


willysr:
merged

PR #1462, opened 2022-04-08
Update mailspring and s6-rc to their latest versions

mumahendras3:
Updated mailspring and s6-rc to their latest versions.

Thanks!


aclemons:
Staged to my branch. Thanks

PR #1465, opened 2022-04-09
nextcloud-desktop: correct tar url

afhpayne:


willysr:
i fixed it on my branch already

PR #1472, opened 2022-04-11
development/jupyter-ipykernel: Update for 6.13.0

isaackwy:
I initially wrote that newer versions of jupyter-ipykernel after 6.10.0 require a newer python-setuptools; however, I just found out that only jupyter-ipykernel 6.11.0 requires python-setuptools >= 60.0.
For reference, Slackware has python-setuptools = 57.5.0.

jupyter-ipykernel versions after 6.12.0 are compatible with setuptools 57.5.0.


willysr:
merged

PR #1481, opened 2022-04-15
python/dulwich: removed, unused and replaced by python3 version

Ythogtha:
I was going to take over this SlackBuild, but today it is not needed by any other SlackBuild anymore.
All packages requiring dulwich are python3, using python3-dulwich, and this is python2.
Hence I think that removing it is the best course of action, there is really no reason to keep python2 stuff longer than they are needed.


willysr:
merged

PR #1482, opened 2022-04-15
system/redis: updated for version 6.2.6, new maintainer

Ythogtha:


willysr:
Did you test this updates with following scripts that depends on redis?
libraries/php-redis
network/openvas-libraries
network/rspamd
system/redis-py
system/trmaid

Ythogtha:
php-redis is obsolete, but it works fine (I launched some tests from its source tree), and it would also work if it was updated to latest 5.3.7 (simple version bump).
rspamd seems to work fine.
redis-py is very obsolete, python2, but still works here with a few tests.
trmaid is broken, but not by redis : it builds python2 (again...), but its feedparser dependency is python3.
openvas-libraries seems broken : I can't download the sources, I don't really know what it stands for, there's no reference to specifically openvas-library in https://www.openvas.org/

So my update of Redis looks OK, however some other slackbuilds may need rework but not because of this update.

Ythogtha:
Should I push some updates here, even though I might not be willing to own maintainership ?
Some stuff here is bound to be abandoned by their maintainer, seeing how its broken and/or obsolete.

willysr:
probably send the notice the mailing list first?

willysr:
any update?

Ythogtha:
I've pushed updates for redis-py and php-redis.
trmaid is not available anymore on pypi, it seems its disappearing ?
openvas should be either removed or someone should find a new maintainer.

Anyway, redis can be merged, because trmaid and openvas are broken but not because of redis.

willysr:
mergeed

PR #1484, opened 2022-04-15
system/smem: updated for version 1.5, python3, new maintainer

Ythogtha:
This project is mostly abandoned (no updates since 2015) but still works fine, either with python2 or python3 so I switched default to python3, because of the optional matplotlib dependency, better used on its python3-matplotlib version.
Besides, python2 must die (eventually).
Also I had to rename the usr/bin/smem file to smem-py because it conflicts with cfitsio package which is stock Slackware, and already have an usr/bin/smem tool.

Note about download and un-taring : they didn't archive properly the 1.5 version, but it is available on their mercurial repository. The links shows a 1.5.tar.bz2 file (and simple wget gets that filename), but with content-disposition the file is properly named smem-1.5.tar.bz2
Hence not knowing (and not really caring) which file people will have, I try them both.


willysr:
you miss the MD5SUM

Ythogtha:
Well spotted, fixed, thanks.

willysr:
merged

PR #1488, opened 2022-04-15
R: update 4.1.3

afhpayne:


academic/R/slack-desc
```diff
@@ -12,8 +12,8 @@ R: R is a language and environment for statistical computing and
 R: graphics. It is a GNU project which is similar to the S language and
 R: environment which was developed at Bell Laboratories (formerly AT&T,
 R: now Lucent Technologies) by John Chambers and colleagues.
-R:
-R:
+R:
```
aclemons, line 15: Trailing whitespace here

academic/R/R.info
```diff
@@ -1,10 +1,10 @@
 PRGNAM="R"
-VERSION="4.1.2"
+VERSION="4.1.3"
 HOMEPAGE="https://www.r-project.org/"
-DOWNLOAD="UNSUPPORTED"
-MD5SUM=""
-DOWNLOAD_x86_64="https://cran.r-project.org/src/base/R-4/R-4.1.2.tar.gz"
-MD5SUM_x86_64="6e28db9d02c6d3dae51a149b8e261ab1"
+DOWNLOAD="https://cran.r-project.org/src/base/R-4/R-4.1.3.tar.gz"
```
aclemons, line 4: This errors on i586 since there is hard check on line 66. Did did you mean to remove DOWNLOAD="UNSUPPORTED"?

afhpayne: I see that error now. No, this because slackbuilds has r-studio legacy which exists only to support 32-bit. Thus lacking a 32-bit build of R makes the legacy studio pointless.

aclemons:
Squashed with #1489 to my branch

academic/R/R.info
aclemons: Thanks, this built fine on 32bit now with #1489

PR #1489, opened 2022-04-15
R: update 4.1.3 32-bit fix

afhpayne:


aclemons:
Squashed with #1488 to my branch

PR #1492, opened 2022-04-17
games/ppsspp: Updated for version 1.12.3.

LinRs:
- update archives of submodule, thanks for fourtysixandtwo's patch.
 - remove ppsspp-lang
 - add SYSTEM_FFMPEG(default off) option
 - default build with: headless, sdl

### DIFF
--- a/games/ppsspp/README
+++ b/games/ppsspp/README
@@ -2,13 +2,18 @@ PPSSPP can run your PSP games on your PC in full HD resolution. It can
 even upscale textures that would otherwise be too blurry as they were
 made for the small screen of the original PSP.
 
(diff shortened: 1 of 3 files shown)


aclemons:
I get an MD5SUM mismatch for zstd-a488ba114ec17ea1054b9057c26a046fc122b3b6.tar.gz. Could you check this again?

Also I think you should leave the sha1 prefixes on the tarballs in your SlackBuild. Otherwise if someone for whatever reason still has the old tarballs in the directory, your glob will match both.

aclemons:
staged to my branch, thanks

PR #1493, opened 2022-04-17
games/nestopia: Updated for version 1.51.1.

LinRs:
- remove libretro build
 - since 1.51, gui depend fltk instead of gtk3


willysr:
merged

PR #1497, opened 2022-04-18
python/pdfarranger: move to office/pdfarranger; adapted doinst.sh

brobr:
Hi Willy, could you please move pdfarranger to the 'office' category; it is a python3-based counterpart of pdfshuffler which lives there as well… Further, the doinst.sh has been adapted to also cater for icons. Thanks a lot.


willysr:
merged

PR #1502, opened 2022-04-19
system/pcem: update to version f4a1ab9.20220418

pghvlaans:
* Update version.
* Build in networking support by default (brings in line with upstream releases).
* `meson` support was dropped; switch the SlackBuild to `cmake`.


willysr:
merged

PR #1512, opened 2022-04-23
libraries/libxdg-basedir: Updated for version 1.2.3.

newHeiko:
Also switched back to most up to date github repository (meaning: No
releases, just tags)
Also removed incorrect homepage line from slack-desc.


aclemons:
Staged to my branch, thanks. Note, I updated the tarball handling so that you get the same filename whether you download with a browser or with wget etc.

PR #1515, opened 2022-04-23
network/signal-desktop: Updated for version 5.40.1

ArTourter:
Same version as the previous PR but fixed absolute symlinks after running sbopkglint


ArTourter:
since the new release was done before this was merge, I'll close it and keep the change for the next version.

PR #1524, opened 2022-04-27
python/python3-aiohttp: Remove gunicorn from DEPs

isaackwy:
gunicorn is not a build or a runtime dependency for python3-aiohttp. 
Rather, gunicorn is an optional dependency - gunicorn is used for deploying the aiohttp web server.


willysr:
merged

PR #1525, opened 2022-04-27
system/pcem*: Move to the desktop category

pghvlaans:
On reflection, pcem and pcem-roms probably fit better in the desktop category since the program is GUI-only.


willysr:
i don't think it's worth to move a script because normally people will use the search or use automatic tools such sbopkg or sbotools to build/install the script. It just give more task for the admins during public update.

pghvlaans:
OK, understood. I'll go ahead and close.

PR #1532, opened 2022-04-28
system/trmaid: removed, upstream disappeared, removed from pypi

Ythogtha:
It may not be my place to remove this, but :
 * source doesn't exist anymore ;
 *  the project seems abandoned for 4 years ;
 *  it has been removed from pypi which was the only place it was available ;
 *  and no other distribution ships it (according to repology which is often right about those things).

I haven't asked Dimitris specifically, but with his 500+ SlackBuilds, I think he has other things to do.
Besides nobody answered on the mailing-list about this, nobody seems to care, which is understandable for a dead project.

Hence I suggest the removal!


willysr:
merged

PR #1540, opened 2022-04-28
misc/dos2unix updated for 7.4.2

bpribula:
dos2unix updated for version 7.4.2

For the slackbuild I updated the copyright and version. Changed the make block to use DESTDIR and removed the instruction to copy over docs since this is taken care of by make. Also, removed INSTALL.txt since this is not needed.


willysr:
merged

PR #1541, opened 2022-04-28
network/signal-desktop: Updated for version 5.41.0

ArTourter:
also fix absolute symbolic link in usr/bin


willysr:
merged

PR #1548, opened 2022-04-29
desktop/dunst: Updated for version 1.8.1

bolokanar:


willysr:
merged
no need to add additional check on tarball since the tarball URL is consistent

PR #1568, opened 2022-05-04
python/python2-yara updated for version 4.2.0 and renamed from yara-python

bgrundy:
- Updated yara-python to 4.2.0 and renamed to python2-yara
- removed yara-python (now python2-yara)


aclemons:
Hey @bgrundy, can you submit this via the submission form - new builds and renames need to go through that process to have the database correctly updated.

PR #1570, opened 2022-05-04
network/onedrive: Udated to version 2.4.17

brobr:


network/onedrive/onedrive.SlackBuild
```diff
@@ -24,7 +24,7 @@
 cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=onedrive
-VERSION=${VERSION:-2.4.16}
+VERSION=${VERSION:-2.4.17}
```
aclemons, line 27: Could you fix the man pages so they are installed with ownership root:root and permissions 0644 please?

```
# stat -c '%a %n' /usr/man/man1/onedrive.1.gz
755 usr/man/man1/onedrive.1.gz
```

brobr:
Hi Andrew,

Thanks for this and the reason why this could happen; I've passed
that information on to upstream. 

Maybe that extra line can be dropped in the next upgrade; wait and see
;-)

Cheers,

Rob

PR #1573, opened 2022-05-05
network/onedrive: Udated to version 2.4.17; chmod 0644 man archive

brobr:
Hi Andrew (@aclemons), hope this does the job. Thanks for spotting this; any idea why the archive ended up as executable?

Cheers, 
Rob


aclemons:
> Hi Andrew (@aclemons), hope this does the job. Thanks for spotting this; any idea why the archive ended up as executable?

Looks like because upstream are not setting any permissions on the invocation of `install` here: https://github.com/abraunegg/onedrive/blob/v2.4.17/Makefile.in#L110

I think that means permissions will default to `rwxr-xr-x`

aclemons:
This is staged to my branch now, thanks

brobr:
Thanks a lot.

PR #1585, opened 2022-05-09
network/zoom-linux: Updated for version 5.10.6.2943

earies:


aclemons:
I've staged this to my branch. Could you check on a couple of nitpicks for the next zoom update?

The `doinst.sh` calls `/usr/bin/gtk-update-icon-cache -f usr/share/icons/hicolor`, but there is nothing in that dir in the package and is thus redundant.

Additionally, the desktop file has a couple of things:

```
key "Encoding" in group "Desktop Entry" is deprecated
value "Network;Application;" for key "Categories" in group "Desktop Entry" contains a deprecated value "Application"
```

Finally, there is a warning from doinst.sh when installing the package:

```
zoom-linux-5.10.6.2943-x86_64-1_SBo: Video conferencing client
xdg-mime: file '/usr/share/mime/packages/zoom-linux.xml' does not exist
```
That is an absolute path, but packages can be installed into another ROOT.

PR #1588, opened 2022-05-13
SlackBuild script for build2.

b8u:
This is a SlackBuild script for Slackware 15.0. Probably I have to merge it to 15.0 branch?


aclemons:
We can only accept new submissions via the webform here: https://slackbuilds.org/submit/ . Could you please submit this there? Thanks

PR #1597, opened 2022-05-19
Fixing slackbuild to work offline

kermitdafrog8:
Fixed issues where the slackbuild was looking for code from a remote site.  Now all code needed for build is downloaded by the script and added that spdlog be a dependency.


aclemons:
staged to my branch, thanks

PR #1603, opened 2022-05-20
gis/gdal: Updated for version 3.5.0

ArTourter:
Shared library .so-version bump
Update build process to cmake (autotools deprecated)
Add option to build Java bindings
Add note for optional dependency for libheif in README


willysr:
Since gdal is being used by many scripts either directly or indirectly, i would like to take more test to confirm that this update doesn't mess up with the following scripts
gis/Fiona
gis/GMT
gis/OTB
gis/grass
gis/merkaartor
gis/openorienteering-mapper
gis/osgEarth
gis/pdal
gis/pktools
gis/postgis
gis/python3-cartopy
gis/qmapshack
gis/rasterio
gis/saga-gis
graphics/paraview

can you check them?

aclemons:
I'm rebuilding all of these at the moment

ArTourter:
I have already tested
- python3-cartopy (one of mine)
- qgis
- GMT
- openorienteering-mapper
- postgis
- Fiona

I am rebuilding on grass and paraview but they are quite big so taking some time.

aclemons:
All those *built* fine for me - I have not run any of them however.

ArTourter:
I have run all the ones I have built. The caveat being that I don't use most of them so they start but I couldn't tell if the features using gdal actually work.

I know that qgis required building gdal without 12bit jpeg support as gdal used to replace the jpeg lib with the internal one for both 12bit and 8bit. They have now changed that and use different libs for 8bit and 12bit so it should no longer be a problem. By default gdal now builds the 12bit lib and with the new version qgis doesn't segfault (but I don't use qgis so I do not know if there was something specific that made it segfault or it would just do it at launch).

aclemons:
This is staged on my branch.

PR #1607, opened 2022-05-22
system/afdko: Updated for version 3.8.3.

duncan-roe:
As advised, had to reinstate downloading antlr4 from internet.


aclemons:
https://github.com/adobe-type-tools/afdko/blob/3.8.3/requirements.txt#L11

Looks like we are missing the dependency on tqdm - and also the version on sbo is now too old for this.

aclemons:
It seems the tqdm maintainer is not active anyway and we bumped that last version ourselves, so I have done that now and merged this change to my branch. thanks.

PR #1608, opened 2022-05-22
academic/sage-bin: Fix .info.

duncan-roe:
Home page URL starts "https" as per https://repology.org/repository/slackbuilds/problems.
Also changed 1st download to start "https://mirrors.mit.edu" which matches network/openntpd but not academic/sage.


willysr:
merged

PR #1620, opened 2022-05-25
Please update testpath and threadpoolctl

isaackwy:
Regarding both testpath and threadpoolctl:
The source tarball no longer provides a default setup.py.

Therefore, my SlackBuild now creates a setup.py shim.
(Otherwise, I would require python-build, flit-core and python-installer as build dependencies).

I based my changes on the python-tomli SlackBuild available within default Slackware repositories:
https://mirrors.slackware.com/slackware/slackware64-15.0/source/l/python-tomli/python-tomli.SlackBuild


willysr:
merged

PR #1622, opened 2022-05-26
network/signal-desktop: Updated for version 5.44.1

ArTourter:


ArTourter:
Sorry for the quick update, upstream released it not long after I submitted my previous PR.

I have based this one back to master. I am still not sure what is best from your point of view, when a PR has already been merge into one of the  admin's branches, whether to update the previous commit, or as I did here, to go back to master and change from there.

aclemons:
I don't have a good answer for that. We had a discussion about this recently, but no concrete outcome. Either is fine for me - it is simple enough to drop the previous commit. We currently don't sync the slackbuilds.org git branches here to github apart from master and 15.0, so you need to add both as a remote in your local repo if you want to base off our branches (and then it will show as a conflict here when you open the PR since you can't set the correct target branch). This is easy to do if you know git, but some people aren't very familiar with git, so it is can be a big ask.

aclemons:
Anyway, I have staged this to my branch now.

PR #1625, opened 2022-05-28
Please update QtPy, terminado, rstudio-desktop and ipython

isaackwy:
terminado no longer uses setup.py by default.
Therefore, the terminado SlackBuild now creates a setup.py shim.


willysr:
merged

PR #1627, opened 2022-05-29
development/tiled: updated for version 1.8.5

Ythogtha:


development/tiled/tiled.SlackBuild
```diff
@@ -27,7 +27,7 @@
 cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=tiled
-VERSION=${VERSION:-1.8.4}
+VERSION=${VERSION:-1.8.5}
 BUILD=${BUILD:-1}
```
aclemons, line 31: This write to /usr/lib on x86_64? Is that expected?

Ythogtha: Oh, looks like the patch is silently broken.
The software works perfectly well, but you're right stuff isn't in its right place.
I thought it was a problem for the upcoming 1.9 release, which changes things, and I was having a hard time fixing it on 1.9-beta, but it's already broken on 1.8.

I will look into it. In the meantime you may, or not, merge that, because I've checked and 1.8.4 already put stuff in /usr/lib/, the breakage isn't new. So at least the software would be up-to-date, and when I find the proper way to fix the lib64 path I'll push an update.

aclemons:
staged to my branch, thanks. If you have a fix for the lib64 thing before the next update, just open another PR and notify me.

PR #1648, opened 2022-06-03
system/fish: Don't clobber printf and time man pages

pghvlaans:
Not the maintainer, but I installed and uninstalled fish a while back, and these two man pages were still present afterwards (`time.1.gz` is from `man-pages` rather than `coreutils`, but I think it belongs here).


willysr:
merged

PR #1649, opened 2022-06-04
Fix indent spacing for terminado, testpath, threadpoolctl and traitlets

isaackwy:
Within the python setup.py shim, each indent should be 4 spaces rather than 6 spaces.


willysr:
merged

PR #1654, opened 2022-06-06
development/jupyterlab_server: Update for 2.14.0

isaackwy:
jupyterlab_server no longer has a setup.py by default - therefore, I added a setup.py shim.


aclemons:
staged to my branch, thanks

PR #1656, opened 2022-06-07
python/traitlets: Update for 5.2.2.post1

isaackwy:
I just found out that jupyter-ipykernel does not build because of traitlets updating to 5.2.2.
Instead, traitlets needs to be updated to 5.2.2.post1.
I also switched the build source to pypi rather than github.


aclemons:
Thanks for this. I had jupyter-ipykernel on my list of builds which were failing after the last public update.

aclemons:
staged to my branch, thanks

aclemons:
@willysr Can you drop this commit from your branch: https://git.slackbuilds.org/slackbuilds/commit/?id=5bb5940b6

PR #1674, opened 2022-06-15
academic/ViennaRNA: Updated to 2.5.1

brobr:


aclemons:
Can you please check the /usr/share/info which is in the final package? This directory is not allowed.

PR #1676, opened 2022-06-16
red/master

RSKYS:
This is assurance layer for future, maybe now idk
Worked fine without that patch in first place but incase someone had problem they can flag that  **RPC=yes**
..

Uh discord also have new version, but I'm not installing it using SBo so I don't really care;
I think I also have two packages to update, but they're not necessary, so I update them later


willysr:
merged

PR #1679, opened 2022-06-17
gis/proj: Updated for version 9.0.1

ArTourter:
- remove mandir patch since it was merged upstream
- add TESTBUILD option to build and run tests (off by default)
- fix /usr/share/doc path in package (bug in CMake definitions)


aclemons:
staged to my branch, thanks

PR #1682, opened 2022-06-19
Update s6 stack

mumahendras3:
Bumped the entire s6 stack SlackBuilds to their latest versions with some general fixups/cleanups, removed unneeded `sed` command in `system/s6/s6.SlackBuild`, and added a new compile-time option in `system/s6-linux-init/s6-linux-init.SlackBuild`.


willysr:
merged

PR #1693, opened 2022-06-25
development/jupyter-ipykernel: Fix SlackBuild setup.py shim

isaackwy:
Someone reported that jupyter-notebook was unable to load jupyter-ipykernel.
(However, I was able to load jupyter-ipykernel in jupyterlab).

I found the error- the setup() line should not contain `setup(name='${PRGNAM}',`
Instead, the setup() line should contain  `setup(name='ipykernel}',`

Also, the added line `import shutil` is due to the command `shutil.rmtree(dest)`


aclemons:
Ok, I've staged this, but instead of hardcoding, I've used your SRCNAM variable which already is set to the name minus the python- prefix.

PR #1719, opened 2022-07-03
python/python3-kiwisolver: Revert to 1.4.1

isaackwy:
1. My recent update to python3-kiwisolver 1.4.3 has broken python3-matplotlib functionality.
python3-matplotlib no longer imports.

2. python3-kiwisolver >= 1.4.2 requires setuptools 61.2. Slackware 15.0 has setuptools 57.5.
Therefore, please revert python3-kiwisolver to 1.4.1 (as it depends on setuptools >= 42).


aclemons:
Addedd `wheel` as a dependency so it builds and staged to my branch, thanks

PR #1743, opened 2022-07-14
network/ircd-hybrid: Updated for version 8.2.40

earies:


aclemons:
`mv: cannot stat 'etc/ircd/ircd.conf.new': No such file or directory`

That file does not exist in the package. There is a `reference.conf`  in `/etc/ircd/`. I've removed that line from `doinst.sh` and this is staged to my branch now.

PR #1744, opened 2022-07-15
network/owncloud-client: Update for 2.10.1.7389

isaackwy:
The following patch is required for building owncloud-client on Slackware-current (it has gcc 12):
https://bugzilla.redhat.com/show_bug.cgi?id=2041135

However, builds in Slackware-15.0 do not require the patch mentioned above.

Ponce, if you're reading this, please add the patch listed above on your slackbuilds current repo.


aclemons:
staged to my branch, thanks

PR #1759, opened 2022-07-22
desktop/qtile: Fix xinitrc

isaackwy:
Another user mentioned that the xinitrc.qtile file is not working properly.

I fixed the xinitrc file (replaced 'qtile' with 'qtile start')

Updated version of qtile have different syntax for starting qtile.


aclemons:
staged to my branch, thanks

PR #1770, opened 2022-07-26
development/jupyterlab_pygments: Update for 0.2.2

isaackwy:
According to setup.py, jupyterlab_pygments 0.2.2 only requires jupyter_packaging as an additional (build) dependency.

However, according to pyproject.toml, jupyterlab_pygments >= 0.2.0 requires jupyterlab as well.

I tried using jupyterlab_pygments and jupyter-notebook without jupyterlab - jupyter-notebook ran without errors.
However, I am not sure whether newer versions of jupyterlab_pygments require jupyterlab as an additional runtime dependency.

For context, jupyterlab in return depends on jupyterlab_pygments (potentially a circular dependency).


aclemons:
I looked through the diff there in the jupyterlab_pygments repo - this could be a copy/paste error with the dependency in pyproject.toml. You've tested it to work, so let's avoid any circular dependency for now.

aclemons:
I've staged this this to my branch now

PR #1782, opened 2022-07-30
development/rstudio-desktop: Update for 2022.07.1+554

isaackwy:
Changes:
- Stripping ELF binaries should be done after building rstudio-desktop with cmake, not before
- Use the gwt from rstudio instead
- Add disable_quarto patch (cmake ignores -DQUARTO_ENABLED=FALSE for some reason)


aclemons:
staged to my branch

PR #1789, opened 2022-08-02
development/jupyterlab: Remove jupyter-notebook requirement from setup.cfg

isaackwy:
For jupyterlab 3.4.4, why did upstream add a jupyter-notebook requirement to setup.cfg?
Line 41 (under the install_requires section) of the source setup.cfg contains the following line:
notebook<7

I thought that because we already have jupyter-notebook_shim, we didn't need jupyter-notebook?

If you would like to check the differences between jupyterlab 3.4.3 and 3.4.4:
diff -r jupyterlab-3.4.3 jupyterlab-3.4.4


aclemons:
staged to my branch thanks

PR #1794, opened 2022-08-05
python/debugpy: Update for 1.6.2

isaackwy:
I would like to keep copyright at 2022 (year when debugpy was uploaded to SlackBuilds.org), rather than 2021-2022 (I created the debugpy SlackBuild script in 2021).


aclemons:
staged to my branch

PR #1862, opened 2022-09-02
gis/proj: Updated for version 9.1.0

ArTourter:


ArTourter:
Note that this package does not bump the shared library .so-version. However I have tested several of the dependees and all worked fine.

PR #1877, opened 2022-09-08
netcdf updates

ArTourter:
I have asked on the list if I could take over netcdf but haven't received any answer so I though I send a PR and use it to open the discussion.

I have tested all the dependees and all worked fine with the exception of grace because they moved the download to https from ftp (commit for this is also included although I do not use it nor do I plan to take it over).

I have an update for cdo ready but will submit when this new netcdf is through. I also have created a package for netcdf4-python module and its dependencies but will also wait before submitting them.


willysr:
merged

PR #1883, opened 2022-09-11
Update s6 stack for the latest version

mumahendras3:
Update skalibs, execline, and s6 for their latest versions.

Thanks!


willysr:
merged

PR #1884, opened 2022-09-11
network/yle-dl: Updated for version 20220830.

maramon:


willysr:
I removed python3-installer as it's second level dep as we have python3-flit_core as first level dep

PR #1885, opened 2022-09-11
python/pikepdf: reverted to 5.3.2 for 15.0 to suit qpdf version

brobr:
python/pikepdf: reverted to 5.3.2 for 15.0 to stay in line with 15.0 qpdf version


willysr:
merged

PR #1898, opened 2022-09-13
games/freeciv: updated for version 3.0.3

Ythogtha:
Installs docs on standard location, and don't duplicate files, make install does a good job.


willysr:
merged

PR #1904, opened 2022-09-14
python/python3-automat: sets m2r as optional dependency

Ythogtha:
python-m2r depends on mistune which can't (yet) easily be updated.
As m2r is not a runtime dependency, but only creates better documentations inside the package, I set it up as optional to avoid forcing people to install an out-of-date and possibly insecure mistune.
This changes nothing for packages using Automat, namely Twisted and stuff relying on Twisted.


willysr:
merged

PR #1927, opened 2022-09-23
desktop/qtile: Update for 0.22.1

isaackwy:
From now on, only add wayland desktop file if pywlroots is installed.


willysr:
merged

PR #1933, opened 2022-09-24
python/python2-attrs: downgraded to version 21.4.0

Ythogtha:
Why did I ever think it'd be a good idea to try and update a python2 package is beyond my today's self understanding.


willysr:
merged

PR #1936, opened 2022-09-26
python/traitlets: Change build method to python3-hatchling

isaackwy:
Someone reported having problems building traitlets using my setup.py shim.
For this reason, I decided to change build method to python3-hatchling instead.


aclemons:
staged to my branch

PR #1937, opened 2022-09-26
development/jupyter_core: Change build method to python3-hatchling

isaackwy:
Build jupyter_core with python3-hatchling instead.


aclemons:
staged to my branch

PR #1938, opened 2022-09-26
development/jupyter_client: Change build method to python3-hatchling

isaackwy:
Build jupyter_client with python3-hatchling instead.


aclemons:
staged to my branch

PR #1969, opened 2022-10-05
perl/perl-Regexp-Common: updated for version 2017060201

NicholasBHubbard:
Update Perl distribution Regexp::Common from v2013031301 circa 2013 to latest v2017060201 circa 2017.


wbraswell:
Looks good from a Perl perspective!

willysr:
merged

PR #1982, opened 2022-10-08
python/argon2-cffi-bindings: Use system argon2

isaackwy:
Also, add commenting regarding the variable PRINT_PACKAGE_NAME


willysr:
merged

PR #1983, opened 2022-10-08
python/argon2-cffi: Change build method to python3-flit_core

isaackwy:
Also, remove the environmental variable ARGON2_CFFI_USE_SYSTEM=1.
That has no effect for this package, unlike with argon2-cffi-bindings.


willysr:
merged

PR #2002, opened 2022-10-13
academic/cdo: Updated for version 2.0.6

ArTourter:


willysr:
i removed hdf5 since it's 2nd level dependency of netcdf

willysr:
still, it's 2nd level dependency, so we need to stick with that rule

PR #2008, opened 2022-10-14
office/pdfarranger: Upgrade to 1.9.1

brobr:


willysr:
i think you miss the simplify_icon_name.patch file
also, make sure to keep the permission at 0644 instead of 0755

PR #2022, opened 2022-10-17
multimedia/kodi: Add patch to fix building w/ newer fmt versions

bassmadrigal:


bassmadrigal:
Shoot, I forgot to update the BUILD number. Can you do that or should I send another commit?

willysr:
is this to solve building issue with stable or current? if this doesn't affect stable, i think no need to bump BUILD

bassmadrigal:
It is to fix a broken build on stable due to the newer fmt in the SBo repo.

willysr:
merged

PR #2025, opened 2022-10-19
python/python3-seaborn: Update for 0.12.1

isaackwy:
scipy3 DEP is now optional.
Build method is now using python3-flit_core.
Due to the seaborn 0.12.0 series being experimental, I waited until seaborn 0.12.1 was released before updating.


willysr:
merged

PR #2030, opened 2022-10-21
audio/ncspot: Updated for version 0.11.0.

pghvlaans:
This is the most recent version of `ncspot` that will build on 15.0 without `rustup`.

Depending on what kinds of features are included in future versions, I might start using `rustup` and submit `ncspot-bin` as an alternative for anyone who doesn't want internet-accessing SlackBuilds. The version will be frozen for the time being, though.


willysr:
merged

PR #2042, opened 2022-10-24
python/python3-editor: Renamed to python3-python-editor

Ythogtha:
This is for the renaming of python3-editor.
Because there are two different python projects : editor and python-editor.
editor's packages should have the names python*-editor, hence python-editor shall get the names python*-python-editor.

This has been brought to my attention on the Repology project, I asked and have been answered there :
https://repology.org/project/python:editor/report

Nothing depends on this package so nothing unusual should happen.
Adding of new python3-python-editor package is being done via the submission form.


willysr:
merged

PR #2085, opened 2022-11-09
python/python3-pandas: Change REQUIRES from numpy3 to python3-numpy

isaackwy:


aclemons:
This is already staged here:
https://git.slackbuilds.org/slackbuilds/log/?h=python-renames

PR #2091, opened 2022-11-10
development/bazel: Update for 5.3.2

isaackwy:
bazel does not run if the binary is stripped - therefore, if you're receiving an error from sbopkglint about usr/bin/bazel not being stripped, please ignore them.


aclemons:
staged to my branch

PR #2092, opened 2022-11-10
python/python3-h5py: Update REQUIRES, remove build optimization

isaackwy:
1. Change numpy3 dependency to python3-numpy
2. Remove build option --optimize=1
3. Remove PKG-INFO from /usr/doc/python3-h5py-3.7.0


aclemons:
staged to my branch. I left out the dependency change since these are all staged on this branch already:
https://git.slackbuilds.org/slackbuilds/log/?h=python-renames

PR #2098, opened 2022-11-12
gis/gdal: Updated for versoin 3.6.0

ArTourter:
removed MYSQL_INCLUDE_DIR invocation as compilation fails with it now


ArTourter:
I have tested most packages that depend on gdal and all have compiled without issues. I have tested the resulting binaries though as I don't use most of them.

willysr:
merged

PR #2127, opened 2022-11-19
academic/root +x README and subdirs

arabusov:
A minor fix. Previously, README was just a file, now (actually, for quite a while) it is a directory. For 5th ROOT version `root.Slackbuild` was setting README's permissions to 644 and this line was just C&P to the current slackbuild, thus nobody except the superuser could actually read the README content.


willysr:
merged

PR #2147, opened 2022-11-24
office/mu: Updated to version 1.8.11

NicholasBHubbard:
Updated office/mu to the most recent release. The autoconf configuration for this release uses an m4 macro that is not available for Slackware 15.0's version of autoconf, so I remove the macro from the configure.ac file before invoking autoreconf.


willysr:
merged

PR #2154, opened 2022-11-27
update and run-time dependency on readme

clavisound:
As the title writes.


willysr:
merged
next time, make a single commit to change everything on 1 script

PR #2169, opened 2022-12-07
development/jupyter-nbconvert: Update for 7.2.6

isaackwy:
jupyter-nbconvert 7.2.6 no longer requires an offline style.css


willysr:
merged

PR #2185, opened 2022-12-13
python/python-executing: Removed (renamed python3-executing)

isaackwy:
I already uploaded python3-executing to SlackBuilds.org.
I was supposed to rename python-executing to python3-executing (as my SlackBuild only depended on Python 3).
However, python-executing is still at SlackBuilds.org.


willysr:
merged

PR #2189, opened 2022-12-16
development/rstudio-desktop: Update for 2022.12.0+353

isaackwy:
The rstudio-desktop build now detects the cmake -DQUARTO_ENABLED option.
I no longer need the disable_quarto patch (specifically for disabling quarto).


willysr:
merged

PR #2194, opened 2022-12-17
python/regex: Remove regex (renamed python2-regex)

isaackwy:
I already split regex into python2-regex and python3-regex.
However, regex is still at SlackBuilds.org.
Please remove regex from SlackBuilds.org (I already replaced it with python2-regex).


willysr:
merged

PR #2195, opened 2022-12-17
python/python-prometheus_client: Removed (renamed python3-prometheus_client)

isaackwy:
I already uploaded python3-prometheus_client to SlackBuilds.org.
I was supposed to rename python-prometheus_client to python3-prometheus_client (as my SlackBuild only depended on Python 3).
However, python-prometheus_client is still at SlackBuilds.org.


willysr:
merged

PR #2202, opened 2022-12-19
network/whalebird: Update for 4.7.3

isaackwy:
Also, fix SlackBuild.
The electron directory is at /usr/share/electron, not /usr/share/electron19


willysr:
merged

PR #2207, opened 2022-12-21
python/pywayland: Add CVE-2007-4559 patch, update README

isaackwy:
1. Security researchers at Trellix (@TrellixVulnTeam) have found a directory traversal vulnerability.
The patch for CVE-2007-4559 addresses that.
More details regarding the patch are available at the link below:
https://github.com/flacjacket/pywayland/pull/42

2. On Slackware 15.0, pywayland does not update to 0.4.15.
This is because the upstream setup.py invokes the following command: `from setuptools.command.build import build`


willysr:
merged

PR #2210, opened 2022-12-21
python/jsonpointer: Update for 2.3 (+new maintainer)

isaackwy:
I tried sending the current maintainer (Alan Aversa) an email asking him to update jsonpointer or allow me to take over maintenance; however, the email did not get sent.
I received an error stating "Recipient address rejected: Domain not found".
I think that the email address no longer exists.
Therefore, I would like to take over jsonpointer.

Also:
1. I made sure that the SlackBuild only builds in Python 3. The jsonpatch SlackBuild (which depends on jsonpointer) only builds in Python 3 as well.
2. Speaking of jsonpatch, there is no need to update the SlackBuild. jsonpatch 1.2.5 (the current version at SlackBuilds.org) depends on jsonpointer >=1.9.


willysr:
merged

PR #2218, opened 2022-12-25
add kernel xanmod-rt

noisecode3:
Hi, I made kernel packages, the idea is that the user have  a local copy of the kernel code to experiment on, set up the configuration with tool like xconfig. Then build with the SlackBuild's it will, look for the config in /usr/src/linux-???-rt?-xanmod1. The scripts are based on how Slackware compile the kernel packages. With only x86_64 config for now.


willysr:
I'm not sure we can accept such submission because kernel-* thing belongs to core Slackware and new submission should be submitted via submission form, not into PR

PR #2222, opened 2022-12-28
system/trash-cli: Add python3-psutil as dependency

isaackwy:
The SlackBuild requires python3-psutil as a build dependency.

Also clean up slack-desc.


aclemons:
staged to my branch, thanks

PR #2247, opened 2023-01-01
Correct missing deps for vdirsyncer

Tonus1:
Couldn't test that on my buid VM without an account to sync.


willysr:
python/python3-aiostream need to be submitted via submission form, not via PR. I think it's best to send the PR after this new package is approved in /ready so that we don't end with a non-existing script.

Tonus1:
On Mon, Jan 02, 2023 at 06:08:23AM -0800, Willy Sudiarto Raharjo wrote:
>python/python3-aiostream need to be submitted via submission form, not
>via PR. I think it's best to send the PR after this new package is
>approved in /ready so that we don't end with a non-existing script.
>

Thanks for clarification. I thought it was the best path since 
vdirsyncer is broken in its current state (those are runtime deps).

I will submit via the web form.

Regards,

Tonus

willysr:
either way, it will be fixed in this week's public update, so both via git or submission form will have to wait until then

PR #2281, opened 2023-01-09
network/vdirsyncer: add new dependencies

Tonus1:
Now that the missing python3-aiostream has been merged, here is the correction of the missing deps for vdirsyncer


willysr:
merged

PR #2371, opened 2023-01-28
system/tbsm: Fix SlackBuild

isaackwy:
1. Build package as noarch.
2. Instead of manually removing /usr/share/doc/tbsm, change build options so that doc folder location is on /usr/doc/tbsm-0.7/doc
3. In particular, fix SlackBuild so that the command `tbsm doc` works.
4. Edit README.SLACKWARE (the list of additional instructions)


willysr:
merged

PR #2373, opened 2023-01-29
python/python3-pytest: Revert to 7.1.3

isaackwy:
1. Under the "NOTE" line, remove "pluggy". The SlackBuild pluggy now only supports Python 2 (as compared with the python3-pluggy SlackBuild).

2. python3-pytest >= 7.2.0 has python3-exceptiongroup as a new dependency.
I cannot build python3-exceptiongroup - it requires python3-flit_scm, which in turn requires python-setuptools_scm >= 6.4.


willysr:
merged

PR #2374, opened 2023-01-29
python/cppy: Update README, SlackBuild format

isaackwy:
1. cppy 1.2.1 requires python-setuptools >= 61.2. I have added info to the README to explain that.
2. Reformat the build script. Only commenting is changed; therefore, bumping build number is unnecessary.


willysr:
merged

PR #2379, opened 2023-01-29
Orphan libcloudproviders and owncloud-client

isaackwy:
I am no longer maintaining libcloudproviders and owncloud-client.
Also, refine SlackBuild format (by removing some commenting).


willysr:
merged

PR #2380, opened 2023-01-30
multimedia/subtitleeditor: Both ``CFLAGS`` and ``CXXFLAGS`` shall be equal to ``SLKCFLAGS``

takusuman:
I was building subtitleeditor right now and perceived that my ``CFLAGS`` settings weren't being in fact used, so checked out the code and figured out that ``CXXFLAGS`` wasn't being set as ``SLKCFLAGS``. Anyway, small but useful fix.


willysr:
merged

PR #2382, opened 2023-01-31
python/jdcal: Orphan maintenance of jdcal

isaackwy:
jdcal is no longer a dependency of python3-openpyxl. I would like to orphan jdcal.
Also, remove some commenting from the SlackBuild.


willysr:
merged

PR #2387, opened 2023-02-02
development/jupyterlab: Edit README

isaackwy:
jupyterlab 3.6.0 has just been released. However, it requires jupyterlab_server 2.19.0.
Slackware 15.0 can only install jupyterlab_server 2.16.3 (any later version of jupyterlab_server requires python-requests >= 2.28. For context Slackware 15.0 has python-requests 2.26.)


willysr:
merged

PR #2394, opened 2023-02-03
python/python3-json5: Update for 0.9.12

isaackwy:
Switch from pypi to github (which contains a more up-to-date version of python3-json5).


willysr:
merged

PR #2410, opened 2023-02-10
python/python3-pytest: Edit README

isaackwy:
Simplify README. Also, the "newer python-setuptools_scm" requirement is only for Slackware 15.0.


aclemons:
staged to my branch, thanks

PR #2432, opened 2023-02-18
development/rust16: Support arm builds.

aclemons:


aclemons:
@pghvlaans Do you mind if we add support for arm?

pghvlaans:
Not at all. This would be for the Slackware ARM project, correct?

I'll actually be pushing the version to 1.66.1 this week (assuming everything that's added rust16 as a dependency still builds), and will incorporate the changes.

willysr:
just FYI, we don't officially support ARM in SBo and some of us don't have machine to test on ARM platform

aclemons:
> Not at all. This would be for the Slackware ARM project, correct?

Yes, in my case to build things for my raspberry pi. Quite a few builds are moving to use this as a dependency now since upstreams are dropping support for older rusts.

> I'll actually be pushing the version to 1.66.1 this week (assuming everything that's added rust16 as a dependency still builds), and will incorporate the changes.

Yes, everything that uses our sort of standard rust template normally builds on arm. I have not found anything that didn't build yet. (except now the things which use `rust16` as a dep don't build since this package does not yet repackage the arm binaries)

PR #2433, opened 2023-02-18
python/python-watchdog: Removed (replaced by python3-watchdog)

isaackwy:
I already replaced python-watchdog with python3-watchdog.


willysr:
merged

PR #2436, opened 2023-02-18
Rename untldr to tinytldr

kovmir:
I am the original maintainer, and author of tinytldr (previously known as untldr). The project has been renamed, and I have changed my e-mail address. Please accept this renaming to the original repository, as the current untldr link is dead.


willysr:
please use the normal submission form for new (or renamed project)
use PR to update or delete from repository only

PR #2440, opened 2023-02-19
Please update python-evdev, pyudev and ds4drv

isaackwy:
1. I am maintaining python-evdev. While Dugan Chen is listed as the maintainer of ds4drv, he had orphaned all of his packages a few years ago.
2. My primary goal is actually to update python-evdev and pyudev. In the process, I have removed python 2 support from both SlackBuilds.
4. I am also updating ds4drv to 0.5.1 - I need to remove Python 3 support from ds4drv.
5. I do not intend to take over maintenance of ds4drv.


willysr:
merged

PR #2442, opened 2023-02-19
python/et_xmlfile: Remove Python 2 support

isaackwy:
The only SlackBuild that depends on et_xmlfile is python3-openpyxl.


willysr:
merged

PR #2444, opened 2023-02-20
python/python3-platformdirs: Edit README

isaackwy:
I can actually update python3-platformdirs within Slackware 15.0.

I am not updating python3-platformdirs though (even though python3-platformdirs 3.0.0 is available upstream).
It's just that if I update it, jupyter_core (and by extension, jupyter-notebook and jupyterlab) breaks.


willysr:
merged

PR #2449, opened 2023-02-24
Filebot and Discord version bumps

bassmadrigal:
network/discord: Version bump to 0.0.25
multimedia/filebot: Version bump to 4.9.6

NOTE: The new filebot version is now paid, so I have uploaded the old, free version as multimedia/filebot-legacy, however, it is in the pending queue.


willysr:
merged

PR #2451, opened 2023-02-25
libraries/tomlplusplus Updated for version 3.3.0

noisecode3:
You right that's to dirty solution to let the use have to build something from current. It was just a matter of this `deprecated: true` syntax... I have tested it with yabridge and the build look the same. Now I "missed the train" yabridge build is now broken for a while :(


noisecode3:
The diff from mesa 1.0.0 and 0.59.4 is

/usr/lib64/cmake/tomlplusplus/tomlplusplusConfigVersion.cmake
if("false") with meson 1.0.0
```
# if the installed project requested no architecture check, don't perform the check
if("") #is now if("false") with meson 1.0.0
  return()
endif()
```
Those changed places now with meson 1.0.0 like this
/usr/lib64/pkgconfig/tomlplusplus.pc
```
includedir=${prefix}/include
libdir=${prefix}/lib64
```
harmless comment in 0.59.4
`####### The input file was /tmp/SBo/tomlplusplus-3.3.0/src/../cmake/tomlplusplusConfig.cmake.meson.in ########`
with 1.0.0
`####### The input file was tomlplusplusConfig.cmake.meson.in ########`

I think those are harmless

willysr:
merged

PR #2478, opened 2023-02-28
games/vcmi: updated for version 1.1.1

Ythogtha:
Also removing packages for mods which are no longer required, or even useful for new version.
Core is included in vcmi, and mods shall be downloaded through the vcmi launcher.


aclemons:
Some database changes here if you could take care of them @willysr

willysr:
it's all good.. database is processed during public update, not when they are in git.
merged

PR #2491, opened 2023-03-03
development/spyder: Update for 5.3.1

isaackwy:
1. I emailed Larry Hajali about a week ago, and he has not answered yet.

2. According to this forum post, Spyder 2.3.9 does not work in Slackware 15.0. This update will make Spyder run in Slackware.
https://www.linuxquestions.org/questions/slackware-14/preferred-python-ide-in-slackware-15-0-a-4175721749/

3. I am not taking over maintenance of spyder because I don't plan on maintaining data science packages for much longer.
That is, in about 1-2 weeks, I will orphan all dependencies related to jupyterlab, jupyter-notebook and spyder.


aclemons:
I think this is fine to update so that it is working. Larry has not responded to multiple emails about other packages. I hope everything is alright.

aclemons:
LGTM - staged to my branch

PR #2533, opened 2023-03-11
system/pcem: Updated for version b8ad941.20230304.

pghvlaans:
Aside from the version bump, I've made two changes:

* Use a wrapper script to make a symlink to `/usr/share/pcem/roms` when running for the first time.
* Build the package even if `/usr/bin/wx-config` belongs to `wxPython`.


willysr:
merged

PR #2536, opened 2023-03-14
Fix setuptools scm pretend version (python3-filelock and python3-virtualenv)

isaackwy:
A user faced issues with the SETUPTOOLS_SCM_PRETEND_VERSION.
I have made the fixes accordingly.


willysr:
merged

PR #2538, opened 2023-03-14
development/rust16: Updated for version 1.67.1.

pghvlaans:
The following packages are currently dependent on rust16:

* bat
* bottom
* clamav
* dust
* fd
* ncspot
* rustup

All of them build under x86_64, i686 and ARM.


willysr:
merged

PR #2554, opened 2023-03-20
development/rstudio-desktop: Update for 2023.03.0+386

isaackwy:
Also, this SlackBuild will download online external node modules.
The README will let users know of that.


aclemons:
LGTM - staged to my branch

PR #2604, opened 2023-03-31
development/bazel: Update for 5.4.0

isaackwy:
1. While bazel 6 is available upstream, bazel 6.x.x does not build with tensorflow (which I plan on uploading later on).

2. The bazel binary should not be stripped - otherwise, bazel does not run.
See issues [#600](https://github.com/bazelbuild/bazel/issues/600) and [#11842](https://github.com/bazelbuild/bazel/issues/11842)


aclemons:
staged to my branch

PR #2613, opened 2023-04-04
system/winetricks: Update for 20230212

isaackwy:
The signature for the upstream winetricks tarball is now good:
```
gpg --verify 20230212.tar.gz.asc
gpg: assuming signed data in `20230212.tar.gz'
gpg: Signature made Mon 06 Mar 2023 11:27:29 AM PST using RSA key ID 053F0749
gpg: Good signature from "Austin English <austinenglish@gmail.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
```


willysr:
merged

PR #2625, opened 2023-04-06
office/kbibtex: Updated for version 0.10.0

ArTourter:
- disabled build testing as requires X which doesn't work with sudo anyway and slows down build
- Using git repo archive download link as the stable kde downloads links are not permanent
- add patch to fix desktop file permissions


willysr:
merged

PR #2633, opened 2023-04-09
system/yabsm: ensure builds with system perl

NicholasBHubbard:
I updated yabsm.SlackBuild to explicitly run yabsm's Makefile.PL with /usr/bin/perl instead of just "perl". If an alternative perl were to come first in $PATH then the shebang of the yabsm executable script will point to this perl. Depending on the version and configuration of that perl, yabsm may run improperly.


willysr:
in a proper Slackware installation system, there won't be any other perl implementation outside the default package so I don't think it's necessary

NicholasBHubbard:
I understand. Thanks for letting me know.

PR #2687, opened 2023-04-14
system/cdemu-daemon: Fix cdemu-daemon udev rule

isaackwy:
A slackbuild user mentioned this message:
`/var/log/syslog: "Apr 14 12:31:34 overlord udevd[731]:  NAME="%k" is ignored, because it breaks kernel supplied names, please remove it from /etc/udev/rules.d/99-vhba.rules:1"`

My change fixes this issue.


willysr:
merged

PR #2726, opened 2023-04-24
python/iniconfig: Remove Python 2 support

isaackwy:
The only package that directly depends on iniconfig is python3-pytest (which is, of course, a Python 3 package).


willysr:
merged

PR #2737, opened 2023-04-27
development/jupyter-notebook_shim: Update for 0.2.3

isaackwy:


aclemons:
This fails sbopkglint because of the directory `/usr/etc`. The previous package also had this, so I'm not going to hold up merging this. Might be worth checking that this really where the file should be or if it is unused.

```
# tree usr/etc/
usr/etc/
└── jupyter
    └── jupyter_server_config.d
        └── notebook_shim.json

2 directories, 1 file
```

staged to my branch

PR #2741, opened 2023-04-27
Wheel rename

fourtysixand2:
DEP changes from wheel -> python3-wheel

Tested against all 249 slackbuilds that had wheel as part of their deptree.

The following python2 builds complain about no python2 wheel but still build successfully and I've left python3-wheel as a dep:
python2-automat
pynacl
rencode


willysr:
merged

PR #2742, opened 2023-04-27
python/python3-tomlkit: Update for 0.11.8

isaackwy:
fourtysixandtwo already uploaded a SlackBuild named python3-wheel (which builds only in Python 3).
wheel (maintained by Dimitris Zlatanidis) builds in both Python 2 and 3.

For now, I am still keeping python3-tomlkit dependent on wheel (until wheel becomes a Python 2 only build).

EDIT: fourtysixandtwo is already requesting a change to python3-tomlkit's DEPs (more specifically, wheel -> python3-wheel):
https://github.com/SlackBuildsOrg/slackbuilds/pull/2741


willysr:
merged

PR #2747, opened 2023-04-28
system/vhba-module: Allow building with custom kernel versions

isaackwy:
This will allow users to rebuild vhba-module after upgrading the kernel (without rebooting).

Another Slackware user made this suggestion.


willysr:
merged

PR #2750, opened 2023-04-28
microsoft-edge: update 112.0.1722.64

afhpayne:
Updating build from original maintainer


willysr:
do you have permission from the original maintainer?

willysr:
merged

PR #2753, opened 2023-04-29
Split pyasn1 and asn1crypto

isaackwy:
Impacket has more dependencies than python2-pyasn1, python2-pycrypto and python3-pyasn1. The real list of dependencies is available at the Arch Linux website: https://archlinux.org/packages/community/any/impacket/

Additional dependencies:
Flask
python3-ldap3 (not available at SlackBuilds.org)
python-pcapy
python3-pycryptodome (on SlackBuilds, this is named as python3-pycryptodomex)
python3-pycryptodomex (not available at SlackBuilds)
pyOpenSSL

However, I am not fixing Impacket (I am not interested in that package).


willysr:
merged

PR #2787, opened 2023-05-07
development/rstudio-desktop: Move /usr/lib to /usr/lib64

isaackwy:
Credits to @slackalaxy.

I now better understand his rstudio-desktop [SlackBuild](https://github.com/slackalaxy/misc/blob/master/preliminary/rstudio-desktop-bin).

slackalaxy extracts the rstudio binary to $TMP, then copies the extracted directories to $PKG (under /usr/lib64).
Therefore, only /usr/lib64 (without /usr/lib) is present in the .tgz package.


willysr:
merged

PR #2815, opened 2023-05-13
Please remove atomicwrites from SlackBuilds.org

isaackwy:
I already uploaded python2-atomicwrites and python3-atomicwrites to SlackBuilds.org.

I already have permission from Dimitris Zlatinidis and Tonus to change the dependencies.
Larry Hajali has not responded - but I don't think he is maintaining SlackBuilds anymore.


willysr:
merged

PR #2839, opened 2023-05-16
updates for my scrips - feh will be submited in another PR, not ready yet...

bolokanar:


willysr:
is there something important in 1.0.3 alpha version?
alpha is usually not stable and it's considered development release

willysr:
i would suggest waiting until it's finally released as 1.0.3, not in alpha state

bolokanar:
Fine.

I'll rework the script and resubmit next week.

PR #2861, opened 2023-05-22
desktop/dwmblocks: Change version formatting

isaackwy:
There isn't a dwmblocks source tarball with the version number 2022.01.07 or 20220107.
Rather, the version number is the date of the git commit.
I would like the verison number to be formatted to 20220107 rather than 2022.01.07 (that is, remove the periods).

Other slackbuilds of this type do not have periods in their version numbers.


willysr:
merged

PR #2862, opened 2023-05-22
network/netcat-openbsd: Updated for version 7.3_1

duncan-roe:
netcat.c knows its version number: no need to put on make line anymore


willysr:
merged

PR #2867, opened 2023-05-24
python/python3-lsp-black: Update for 1.3.0

isaackwy:
python3-lsp-black now depends on python-tomli (available in Slackware 15.0) rather than python-toml (available in SlackBuilds).
From the upstream setup.cfg:
```
[options]
packages = find:
install_requires =
    python-lsp-server>=1.4.0
    black>=22.3.0
    tomli; python_version<'3.11'
python_requires = >= 3.7
```


willysr:
merged

PR #2873, opened 2023-05-26
Python renames

fourtysixand2:
Update deps for renames of pyqt-distutils and docopt.

Changes were tested successfully.


willysr:
merged

PR #2874, opened 2023-05-26
network/teamviewer: Updated for version 15.42.4.

ArTourter:
if the `libicudata.so.*` file provided in the `.deb` package is stripped (even with the `--strip-unneeded` flag on) the program doesn't start and leave the error:

`/opt/teamviewer/tv_bin/TeamViewer: error while loading shared libraries: libicudata.so.63: ELF load command address/offset not page-aligned`

with that file left unstripped, the program works as expected.


willysr:
merged

PR #2877, opened 2023-05-27
development/jupyter_server: Update for 2.6.0

isaackwy:
I have added python3-overrides as a dependency. This is necessary for the update.


willysr:
merged

PR #2879, opened 2023-05-27
python/python3-anyio: Update README

isaackwy:
1. python3-anyio 3.7.0 has just been released; however, it depends on python-setuptools >=64 and python-setuptools_scm >= 6.4.
2. I have also removed some commenting (to fix formatting).


willysr:
merged

PR #2880, opened 2023-05-27
desktop/breath-gtk-theme: Change version format

isaackwy:
This is not actually a version update.
I have only changed the SlackBuild version format.

The Arch Linux AUR similarly lists breath-gtk-theme at version 5.9.0:
https://aur.archlinux.org/packages/breath-gtk-theme

I don't recommend breath-gtk-theme for Slackware 15.0 anyway.
It's an old package meant for GTK 3.16 (for context, Slackware 15.0 has GTK 3.24).


willysr:
merged

PR #2891, opened 2023-06-02
gis/gdal: Updated for version 3.7.0.

ArTourter:


ArTourter:
I have tested most dependees and every thing seems to compile with it. There is nothing in the changelog that imply an incompatible change anyway.

willysr:
merged

PR #2895, opened 2023-06-03
python/iniconfig: Remove iniconfig

isaackwy:
I already have uploaded python3-iniconfig to SlackBuilds.org (as a rename of iniconfig).
Therefore, I would like iniconfig to be removed from the website.


willysr:
merged

PR #2897, opened 2023-06-04
system/FontAwesome: Update for 6.4.0

isaackwy:
Last week, I emailed Andre Fernando (the original maintainer) to update FontAwesome to 6.4.0.
Since the update is more complicated than a version bump, I gave him a tarball for updating FontAwesome.
However, he has not responded (either with updating the package himself or allowing me to make the update on his behalf).

I would like FontAwesome to be updated to 6.4.0.
I am not looking to take over maintenance of FontAwesome. This is my last week of maintaining Slackbuilds.


willysr:
merged
i think the maintainer is no longer active so it's better to take over maintainership if you are using this package

PR #2908, opened 2023-06-09
python/python3-dulwich: fix wrong build using updated setuptools

Ythogtha:
Following discussion on SBo mailing list, this update will fix the build when python3-setuptools-opt is added by fourtysixandtwo.
Please wait until submission, and acceptance, of python3-setuptools-opt, then this should be ok immediately.

There is no need to be smart about the hardcoded python version in PYTHONPATH, because python will be updated on Slackware current, which already have a sufficiently uptodate setuptools. So this is for slackware-15.0 only, hence python 3.9 only.

Furthermore, the PYTHONPATH line does not prevent, in the absence of the given PATH, the proper build of this package on box with an updated setuptools like on slackware-current.
So this'll work everywhere.


willysr:
thanks, will check it out later

willysr:
merged

PR #2909, opened 2023-06-09
python/python3-aiorpcX-legacy: Remove python3-aiorpcX-legacy

isaackwy:
python3-aiorpcX-legacy is no longer a required dependency for Electrum.
nomnombtc uploaded this a few years ago when I incorrectly updated python3-aiorpcX past version 0.18.*.

Now, Electrum depends on python3-aiorpcX (currently at version 0.22.1).

Please remove python3-aiorpcX-legacy from SlackBuilds.org.
nomnombtc has provided me permission to do so.


willysr:
merged

PR #2919, opened 2023-06-13
development/rust16: Updated for version 1.69.0.

pghvlaans:
rust16 currently has the following reverse dependencies, which built successfully on x86_64, i686 and ARM:

* alacritty
* bat
* bottom
* clamav
* dust
* fd
* hyperfine
* ncspot
* newsboat
* rustup
* skim


willysr:
merged

PR #2923, opened 2023-06-14
python/python3-service-identity: updated for version 23.1.0

Ythogtha:
Python build process is becoming even more complicated...
But well, it works on a clean environment :)


willysr:
merged

PR #2924, opened 2023-06-14
python3-celery, python3-kombu, python3-click-didyoumean

Ythogtha:
This is an update of python3-celery.

click-didyoumean was very outdated, and only required by celery, I asked Pr. Horstmann and he gave me ownership.
So here we have :
* a renaming of click-didyoumean to python3-click-didyoumean
* update from 0.0.3 to 0.3.0
* new maintainer (me)
* update of python3-kombu and python3-celery to 5.3.0, kombu is required by celery, and only it, so it makes sense to regroup here also
* update of python3-celery REQUIRES to account for new name of python3-click-didyoumean, and other changes
* cleanup of python3-celery versions patch, because it is not needed anymore

Kallithea, which is the thing that requires all of this, works pretty fine, so I guess this is good :)


willysr:
for renaming, it's better to submit PR for removal and submit new name using submission form because it needs to be populated into the DB

Ythogtha:
Ok, so I submitted python3-clickdidyoumean on the website, and changed the corresponding commit here for removal.
celery REQUIRES the new python3-clickdidyoumean, so this PR needs to wait for pending approval.

willysr:
merged

PR #2927, opened 2023-06-14
ham/hackrf: Updated for version 2023.01.1 + new maintainer.

pyllyukko:
Updated and taking maintenance over.

I asked permission from the original author and here's his response:

> Yes, it's been a long since I have used it and I don't plan to use it in the future.                                                                          
>                                                                                                                                                               
> Please go ahead and take over.

Also changed this into more suitable category.


willysr:
merged

PR #2932, opened 2023-06-15
brave-browser: update 1.52.126

afhpayne:


willysr:
brave browser is already at 1.52.126 so this patch is not based on latest master

afhpayne:
> brave browser is already at 1.52.126 so this patch is not based on latest master

@willsr not sure what you mean. The current SBo is 1.52.117 and this patch is for 1.52.126.

willysr:
i think i merged this PR before and forgot to close it

PR #2946, opened 2023-06-19
Pypies: updated for version 7.3.12

Ythogtha:
All versions of pypy are updated to 7.3.12.
pypy : repackage of binaries targetting python2.7
PyPy : building pypy for python2.7 from source
pypy3 : repackage of binaries targetting python 3.9
PyPy3: building pypy for python 3.9 from source


willysr:
merged

PR #2948, opened 2023-06-19
python/python3-black: reverted to version 22.12.0, for outdated packaging

Ythogtha:
I've looked into it, and even though black builds and works fine (and the same) with stock slackware's python-packaging, project importing it are failing because of the outdated packaging.
I didn't go that far with tesing spyder, because it looked broken with some missing dependencies (namely python3-textdistance, jupyter-qtconsole, QDarkStyle, ipython_genutils, and again for python3-lsp-server : python3-rope, python3-pydocstyle).

But when they are resolved, the updated black still breaks spyder.
I tried having a python3-packaging-opt like the new python3-setuptools-opt, but it turns out it's not a build dependency, but a runtime one, used for import by external projects (like spyder).
I guess tweaking more into spyder just for having a slightly updated black may not be worth it, so I'm reverting to 22.12.0, as it looks like the better solution.

Sorry for not having tested fully, I guess I was going too fast with the number of updates, and new packages last week...


willysr:
merged

PR #2949, opened 2023-06-19
desktop/rss-guard: Removing unused patch files

Ythogtha:
Looks like I forgot to commit the removal of those unused patch files from rss-guard 3.x.
Sorry.
Nothing changes on the build itself, no need to bump build number.


willysr:
merged

PR #2950, opened 2023-06-19
system/hplip-plugin: updated for version 3.20.5. New maintainer.

Ythogtha:
Unmaintained since 2016, maintainer inactive since 2020.
This only needs to be in sync with stock slackware's hplip package version.
I'm using it for my printer.
It's only a version bump.

So I'm taking over, see you on slackware 15.1 for an update ;)


willysr:
merged

PR #2952, opened 2023-06-20
multiple maintainer updates: isaacyu -> nomnombtc

nomnombtc:
Hello,

Electrum has some dependencies maintained by Isaac Yu, who will not update them anymore (see mailing list).
I would like to claim these SlackBuilds, which don't seem to have been taken over by anyone yet:

development/QtPy
libraries/QDarkStyle
python/python3-aiohttp-socks
python/python3-aiorpcX
python/python3-helpdev
python/python3-pypng
python/python3-socks

There are other Electrum dependencies, but as far as I can tell they have already found a new home in
the replies to his mailing list post.

For now, I've only updated the *.info files, I'll update the script copyright when there is an actual 
version/other change needed.


willysr:
please make one commit per script, not a unified changes like this

PR #2953, opened 2023-06-20
games/Mindustry: updated for version 145.1

Ythogtha:


willysr:
is this only for x86_64?
if yes, you should put the tarball into DOWNLOAD_x86_64

Ythogtha:
No, no, it works fine in ix86, it's my mistake, second time the same, but I fixed properly in my files.
I re-pushed the right .info
Sorry !

PR #2993, opened 2023-07-03
games/ags: updated for version 3.6.0.50, libraries/SDL2_sound: updated for version 2.0.2

Ythogtha:
No other package than ags depends on SDL2_sound.


willysr:
merged

PR #2994, opened 2023-07-04
Cleanup python2

Ythogtha:
Following the mail from two month ago, and the new mail from today, 31 packages removed.
Those are unmaintained, obsolete, out of date, python2, well, a lot of serious defect.
And apart from requirements (like python2-twisted), end-packages are not working because their requirements are a mess of python2/python3.

What matters here already exists on python3, this is all garbage.


willysr:
has this been checked with other scripts that depends on one of these?

Ythogtha:
> has this been checked with other scripts that depends on one of these?

```bash
$ pkgs="ZConfig python2-pyhamcrest python2-mistune zdaemon simplegeneric ZODB3 service_identity poster Electrum-DASH python2-pyzmq salt python2-ldap python2-automat gaphas etk.docking python2-twisted txrequests python-m2r python2-klein sslstrip gaphor Parsley python2-orbited graphite-carbon graphite-web txsocksx python2-txaio txi2p python2-autobahn Nevow txtorcon TaskCoach Scrapy foolscap tahoe-lafs"
$ [ "$pkg" = "$(hoorex $pkg)" ] && echo "OK"
OK
```

hoorex tells me that no package outside of that group of packages depends on any package within that group.
So we should be safe here.

willysr:
merged

PR #2995, opened 2023-07-04
Zope takeover

Ythogtha:
For zope.interface only : dropping python2 support.
For all others : updated to latest version, changed maintainer, python3 only, and harmonization of all SlackBuilds files.

/!\ zope.sqlalchemy depends on python3-transaction which I've just added through the submission form.

Stuff that depends on those packages are :
* locust depends on zope.event, and seems to work fine with newer version.
* letsencrypt depends on zope.event, zope.hookable and zope.component, and seems also to work fine with the new version.
* gaphor depends on some of these packages, but it is python2, unmaintained and broken, I'm removing it.
* ZODB3 is a work in progress, it is to be dropped, but replaced by an updated ZODB. Anyway, it is broken now, so it'll stay there and broken until I finalize ZODB.


willysr:
merged

PR #3006, opened 2023-07-09
python/python3-ZODB: fixed

Ythogtha:
Looks like the python3-BTrees is already in REQUIRES, so only a build bump here.


willysr:
merged

PR #3014, opened 2023-07-13
development/rust16: Updated for version 1.70.0.

pghvlaans:
All reverse dependencies build on x86_64, i686 and ARM.

I'm aware of the 20-arch `sbopkglint` failure on 64-bit. Stripping the libraries unfortunately causes memory faults on Slackware64-15.0. Sorry for any inconvenience. The plan at this point is to remain in sync with -current until some version of `rust` fixes this problem (won't be 1.71.0; already checked).


willysr:
merged

PR #3043, opened 2023-07-22
academic/sage: Updated for version 10.0

cduston44:
Still trying to make this work...rebased my master with upstream. Shows only commit difference is this one, hope it works....


willysr:
can you fix the commit to use academic/sage: Updated for version 10.0 ?

cduston44:
> can you fix the commit to use academic/sage: Updated for version 10.0 ?

I changed the title, if that's what you're after.

willysr:
the commit message, not the title :)

cduston44:
Ok, it says that now!

willysr:
@aclemons can you review this PR?

aclemons:
Ok, this took me a couple of tries, but I built it on both i586 and x86_64.

I did not notice that the files in the package don't all have the correct owner and group. Some are root:root but others are 1000:1000 and still others are 501:games. This is not an issue with this change. The version of the script in the repo also produces a package like this. Could you look into updating this for the next time you submit an update?

This change is now staged on my branch.

cduston44:
I will make a note of that, and try my best to understand why and change it if possible.

Sent with [Proton Mail](https://proton.me/) secure email.

------- Original Message -------

PR #3081, opened 2023-08-05
system/bottom: Use static.crates.io for download URLs.

pghvlaans:
I shot an email to the mailing list earlier this afternoon.

Basically, Rust crates with a `+` aren't downloading properly from the amazonaws site. static.crates.io is still working (and it seems much faster), so I'm changing my download URLs over.


aclemons (approved):
LGTM

PR #3175, opened 2023-09-01
python/python3-poetry-core: Updated for version 1.7.0.

fourtysixand2:
No issues testing builds that depend on python3-poetry-core


aclemons (approved):
LGTM

PR #3241, opened 2023-09-18
academic/ViennaRNA: Upgrade to version 2.6.3; new module RNAxplorer i…

brobr:
…nactivated by default (see ViennaRNA.SlackBuild)


willysr:
merged

PR #3244, opened 2023-09-18
development/textadept: updated for version 12.1, new maintainer, building from sources

Ythogtha:
I put some work into that one:
 * taking ownership (that's the easy part :smiley:).
 * building from source.
 * downloading third party sources which otherwise would be downloaded during build - I hope this builds fully offline now
 * making a helper script for future management of this mess.
 * building HTML version of markdown documentation, since the local HTML files are used byt the application itself to open things from the help menu - that wasn't easy, I had to hit the code pretty hard to have something functional.
 * package all that nicely...
 
Let's hope it'll pass all the checks ;)

### DIFF
--- a/development/textadept/README
+++ b/development/textadept/README
@@ -6,14 +6,20 @@ speed and minimalism over the years, Textadept is an ideal editor for
 programmers who want endless extensibility without sacrificing speed
 or succumbing to code bloat and featuritis.
 
-The following executables are symlinked from /opt/ to /usr/bin/:
+The following executables are symlinked from /usr/share/textadept/
+to /usr/bin/:
 
   -textadept
+  -textadept-gtk
   -textadept-curses
-  -textadeptjit
-  -textadeptjit-curses
   -tad
   -tadc
+  -tadg
 
-Note that `tad` and `tadc` are just easier ways to type `textadept`
-and `textadept-curses`, respectively.
+Note that `tad`, `tadg` and `tadc` are just easier ways to type
+`textadept`, `textadept-gtk` and `textadept-curses`, respectively.
+
+python3-markdown-it-py is only required for building HTML
+documentation, and the scripts works fine without it.
+But since the local HTML pages are opened when using the menu/help,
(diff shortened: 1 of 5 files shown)


aclemons:
@sbo-bot: build development/textadept

Ythogtha:
Found the problem, re-pushing...

Ythogtha:
@sbo-bot: obey my command, you machine !

*sigh* doesn't listen to me apparently, might be because of my awful french accent...

aclemons:
@sbo-bot: build development/textadept

Ythogtha:
Ah, I've been able to reproduce that error, and fix it.
Well, instead of adding lua as build-time requirement, since textadept takes great length to not depend on installed lua but to ship its own, I build the internal lua, and uses it for building docs.

Should be ok, :crossed_fingers:

aclemons:
@sbo-bot: build development/textadept

aclemons:
Ok, that built, but looks like it is not using DESTDIR etc for the install?

Ythogtha:
Well, the build/install process really wants to make a bundle.
It is using the /usr prefix, but putting everything in /usr/share/textadept.
So I used symbolic links for binaries and desktop files.
Maybe I should mv them.
I'll try that and if the resulting package works as expected, I'll push this change.

Ythogtha:
Well, after testing, it turns out that binaries expect to be in the same repository as other data like init.lua.
So using symbolic links in /usr/bin is the way to go.
I'll keep it as-is, I don't want to alter too much how textadept does things, we might run into bad side-effects.

Previous script was using pre-built binaries in /opt/textadept/ and symbolic links in /usr/bin/
I like mine better because it builds from sources, even if complexity follows ^^

aclemons:
Ok merging this - the cmake install puts some symlinks on the build host, but they don't actually point to anything. The rest of the stuff goes into the package as it should, so it is mostly ok, just puts some rubbish on the build host.

aclemons (approved):
LGTM

PR #3260, opened 2023-09-23
network/signal-desktop: Updated for version 6.31.0.

ArTourter:
this updates on top of the last release and should replace the latest accepted version as a new version was released since.


willysr:
merged

PR #3262, opened 2023-09-23
Fix readme and info for swift

lockywolf:
Fix README as suggested by the feedback email.


willysr:
README is normally made short, just to give brief information about the project is all about or something they need to consider, such as optional dependencies or warning if it requires internet connection or downloading something during process.

Details about the project should be in the original project website

willysr:
merged

PR #3273, opened 2023-09-26
games/planetblupi: updated for version 1.15.0

Ythogtha:
This version requires a newer SDL2_tff lib than the one provided by slackware.
It allows for more fonts to be used (like right to left fonts), hence is useful for internationalization.
So we are building and shipping our own SDL2_ttf directly - its 95k so not a huge overhead.
Also binary is now installed in /usr/games and not anymore in /usr/bin, as is tradition in slackware.

The whole SDL2_ttf block in planetblupi.SlackBuild could be removed directly for using system's SDL2_ttf, for example when 15.1 is out, and the planetblupi binary will be directly in /usr/games, hence the redefinition of BINDIR.

### DIFF
--- a/games/planetblupi/planetblupi.SlackBuild
+++ b/games/planetblupi/planetblupi.SlackBuild
@@ -23,10 +23,11 @@
 cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=planetblupi
-VERSION=${VERSION:-1.14.2}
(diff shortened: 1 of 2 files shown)


willysr:
@sbo-bot: build games/plateblupi

willysr:
@sbo-bot: build games/planetblupi

willysr:
merged

PR #3292, opened 2023-09-30
system/mpollux-digisign-client: Updated for version 4.2.6.

pyllyukko:


pyllyukko:
Linter complains, but there are different versions for `x86` and `x86_64`. We could also do it as:

```
VERSION=${VERSION:-4.2.6b_8414}

...

if [ "${ARCH}" = "i586" ]
then
  VERSION=4.2.6_8401
fi
```

So it would "default" to `x86_64`.

But that's not optimal either.

willysr:
that's strange to release similar name with just 1 character difference

willysr:
> Linter complains, but there are different versions for `x86` and `x86_64`. We could also do it as:
> 
> ```
> VERSION=${VERSION:-4.2.6b_8414}
> 
> ...
> 
> if [ "${ARCH}" = "i586" ]
> then
>   VERSION=4.2.6_8401
> fi
> ```
> 
> So it would "default" to `x86_64`.
> 
> But that's not optimal either.

this is better solution i guess, we can assume the default is people are using x86_64

willysr:
can the VERSION in .info be adjusted so that it has same value with VERSION in x86_64 arch?

aclemons:
Does upstream every release multiple versions with the same triplet but different suffix? If not, maybe the version of the slackware package can just be the same for both architectures, but the script uses the source version for each arch to extract?

willysr:
since no revision since 2 weeks ago, i'll close this one

PR #3294, opened 2023-09-30
pypy/PyPY/pypy3/PyPy3

Ythogtha:


willysr:
i think the commit message is incorrectly typed twice

Ythogtha:
> i think the commit message is incorrectly typed twice

Yeah, that's a... bug :roll_eyes: 

Fixed !

Ythogtha:
Sorry for all the repush, I was trying to determine if the `python package.py ...` line would work with python3 or required python2, to remove the warning about using unspecified `python` command.
It turns out that for PyPy it requires python2, but for PyPy3 python3 is ok, however it doesn't package the debug stuff.
So I have my answer : python2 is still required for that part also.

I'm fully OK for a merge now, all tested and fine on my part!

willysr:
merged

PR #3317, opened 2023-10-09
libraries/DateTime: Removed. Unmaintained - python2 build with python3 dependencies.

Ythogtha:
I didn't ask the maintainer:
* this build is 6 years old.
* it builds for python2 only.
* it has python3 only dependencies.
It obviously can't work properly as it is, and is also obviously unmaintained.

Let Python2 die in peace.


willysr:
merged

PR #3322, opened 2023-10-09
python/python3-twisted: fixed for zope.interface 6.x

Ythogtha:
zope.interface changed internal name from zope-interface, which breaks twisted dependency checks, but not immediately, hence I didn't see it early enough.
That fixes it.


willysr:
is this x86_64 spesific?

Ythogtha:
After some investigations, it seems to boils down to outdated setuptools.
So the workaround in this PR should be enough for now, until either Slackware 15.1/16.0, or more breakage and usage of setuptools-opt as requirement.
I'm good with the built as it is now.

willysr:
merged

PR #3353, opened 2023-10-13
gis/eccodes: Updated for version 2.32.0.

ArTourter:
removed 32bit support again as even prior to version 2.30.0, 32bit compilation failed.


willysr:
merged

PR #3378, opened 2023-10-17
No dependees and duplicate, use python3-fonttools

lecramyajiv:
As part of the renaming process  I had renamed the python-fonttools
 to python3-fonttools which is already in slackbuilds. Since no packages 
are dependent on python-fonttools we can safely remove it.

Regards
Vijay Marcel


willysr:
merged

PR #3575, opened 2023-11-04
system/fd: Updated for version 8.7.1.

aclemons:


aclemons:
@sbo-bot: build system/fd

PR #3580, opened 2023-11-05
python/josepy: Updated for version 1.14.0.

willysr:


willysr:
@sbo-bot: build python/josepy

PR #3613, opened 2023-11-08
python/python3-twisted: REQUIRES update

Ythogtha:
For acknowledging the renaming of hyperlink to python3-hyperlink, and incremental to python3-incremental.
Do not merge before validation of python3-hyperlink and python3-incremental submission.

Also, sorry about not making only one PR for Twisted, I don't know if this one will be OK, as it also includes stuff from the previous one, or if I need to wait for after the next update.


willysr:
merged

PR #3702, opened 2023-11-14
desktop/nwg-shell-config: Updated for version 0.5.24

mac-a-r0ni:
I've also updated the Slackware logo svg to a better suited image that scales properly in the gui.


willysr:
merged

PR #3861, opened 2023-11-26
python/python3-gensim: Update README

isaackwy:
The SlackBuild README is outdated. Cython is now at 0.29.36. Therefore, python3-gensim can be updated to later versions.

I am not returning to regular SlackBuild maintenance just yet (This means I am not updating the python3-gensim version).


willysr:
merged

PR #3970, opened 2023-12-03
Enable SVG support in wxWidgets

spillner:
wxWidgets.SlackBuild has historically forced -DwxUSE_NANOSVG=OFF.  Now that there's a new libraries/nanosvg package on SBo, it makes sense to auto-detect the NanoSVG library and build wxWidgets with SVG support if present.  You can set NANOSVG=no in the SlackBuild environment to disable this behavior, NANOSVG=yes to always include SVG support (using wxWidgets' built-in version of the NanoSVG library if the system package is not present), or NANOSVG=builtin to always use the built-in version.

Several other SBo packages, including games/SLADE, will fail to build if wxWidgets was compiled without SVG support.


willysr:
@sbo-bot: build libraries/wxWidgets

PR #4102, opened 2023-12-10
academic/rpy2: replace install setup.py

brobr:
with thanks to fourtysixandtwo.


willysr:
@brobr : i think you need to add python3-setuptools-scm-opt if you want to use newer setuptools

willysr:
@sbo-bot: build academic/rpy2

willysr:
merged

PR #4141, opened 2023-12-14
Python3 twisted

Ythogtha:
I thought hyperlink and incremental had been removed, replaced by python3-hyperlink and python3-incremental, but it looks like I was wrong.
This PR updates python3-twisted to use the new package names, and removes them.


willysr:
merged

PR #4197, opened 2023-12-20
libraries/gupnp-av add patch for libxml2

0xBOBF:
The current gupnp-av build fails to build since the latest updates to libxml2 in Slackware 15.0. This patch fixes building gupnp-av with the latest versions of libxml2.


willysr:
@sbo-bot: build libraries/gupnp-av

willysr:
merged

PR #3266, opened 2023-09-24
Updated for 10.1, and modified some user and group settings, as requested

cduston44:

PR #4198, opened 2023-12-20
Excess pull request

cduston44:
I don't really understand what happened, but I thought this pull request was approved some months ago. The changes appear upstream, and they were discussed #3266. Can we merge so that I can continue to work?


cduston44:
Ok, we got that error before - the build directory should remain after installation, and the recommendation from the developers is that "sage should not be built as root". If someone could just advise me here - should I try to build it in tmp (as required by sage) and then move is to usr after build is done? And hope I get the path right?

willysr:
this patch doesn't work as we already have sage at 10.1 https://slackbuilds.org/repository/15.0/academic/sage/
@cduston44

PR #4216, opened 2023-12-21
desktop/gnome-session: Update SlackBuild

0xBOBF:
This update removes a couple duplicate files that were installed by the gnome-session package. Also added aarch64 support to the SlackBuild, in line with the SBo templates.


willysr:
merged

PR #4231, opened 2023-12-23
system/android-file-transfer: Update for 4.3

isaackwy:
I plan on orphaning this package at the beginning next year (that is, after this week's update cycle).

Also, remove mentions of qt4 and qt5 in the README.
Slackware 15.0 uses qt5 by default (meaning there is no need to specifically install qt5 as a SlackBuild, unlike with Slackware 14.2).


willysr:
@sbo-bot: build system/android-file-transfer

PR #4258, opened 2023-12-24
system/pass-import: Update README (list python3-pykeepass as optional DEP)

isaackwy:
python3-pykeepass has been approved for uploading to SlackBuilds.org.
Therefore, I am listing python3-pykeepass as an optional dependency in the README.
Also, I am bumping the year in the .SlackBuild file.


willysr:
merged

PR #4263, opened 2023-12-25
multimedia/stremio: Updated for version 4.4.165

noisecode3:


willysr:
@sbo-bot: build multimedia/stremio

willysr:
@noisecode3 : can you check this warning?
```
which: no node in (/usr/libexec/repo:/usr/local/sbin:/usr/sbin:/sbin:/usr/local/bin:/usr/bin:/bin:/usr/games:/usr/lib/libexec/kf5:/usr/lib/qt5/bin)
ln: failed to create symbolic link '/usr/share/stremio/node' -> '': No such file or directory
WARNING: multimedia/stremio: Possible error message from doinst.sh or installpkg
```

noisecode3:
it means it installed without node package, looks okej to me 
```which node```
 ```/usr/bin/node``` 
 should be the output if you have node installed and its listed.

**Comment by noisecode3** [comment]
lol let me make a nicer error there

**Comment by willysr** [comment]
@sbo-bot build multimedia/stremio

**Comment by willysr** [comment]
@sbo-bot: build multimedia/stremio

**Comment by noisecode3** [comment]
Its fixed, can you try manually. somethings wrong cus it's working on my end can you check ?

**Comment by noisecode3** [comment]
@sbo-bot: build multimedia/stremio

File: multimedia/stremio/doinst.sh [hunk 1]
```diff
@@ -6,4 +6,9 @@ if [ -e usr/share/icons/hicolor/icon-theme.cache ]; then
     /usr/bin/gtk-update-icon-cache -f usr/share/icons/hicolor >/dev/null 2>&1
   fi
 fi
-ln -sf "$(which node)" "/usr/share/stremio/node"
+which_node=$(which node 2>/dev/null)
```
aclemons, line 9: Shouldn't this be without the leading `/`? The error output we are seeing in the build output is because the package is being installed into a chroot with a `--root` flag.

willysr:
merged

multimedia/stremio/doinst.sh
noisecode3: thanks for feedback, when the script is used without chroot it looks for where node it installed "At installed time" it makes it work if the user have installed node at any path. but with chroot and without node in there. What wont work. now it will default to /use/share/ anyway. Yes that should be with a leading / its forcing the link there.

PR #4264, opened 2023-12-25
network/broadcom-wl: Updated doinst.sh

mac-a-r0ni:
Received a report about errors with the doinst.sh, the backslashes were causing the $NEW conf file to not be found & renamed properly. Cleaned up the file and verified it now works as intended.


willysr:
merged

PR #4306, opened 2023-12-30
python/python3-shtab: Orphan maintenance of python3-shtab

isaackwy:
I am not interested in bash/zsh completions for trash-cli (which python3-shtab provides).

More importantly:
I don't know how to integrate bash/zsh completion support into a first-time installation of trash-cli.
I have to install trash-cli first (without bash/zsh completion) and then (after installing python3-shtab) re-compile/re-install trash-cli again.


willysr:
merged

PR #4309, opened 2023-12-30
desktop/qtile: Update README and slack-desc

isaackwy:
B. Watson asked me (through the slackbuilds-users mailing list) to set a better description in the README and in the slack-desc:
https://lists.slackbuilds.org/pipermail/slackbuilds-users/2023-December/028811.html


willysr:
merged

PR #4311, opened 2023-12-30
python/terminado: Update for 0.18.0

isaackwy:
I just found out that terminado is a dependency of jupyter_server_terminals, which in return is a dependency of jupyter_server.
If I found out earlier, then I would have included this in the jupyter_server branch, rather than creating a seperate branch just for updating terminado.


willysr:
@sbo-bot: build python/terminado

willysr:
merged

PR #4341, opened 2023-12-31
Libfreeaptx cleaned

thnkman:
OK, i feel as if i have been spamming the pull requests today, but it has been a fun learning process :-)

Happy New Year!

### DIFF
--- a/audio/libfreeaptx/libfreeaptx.SlackBuild
+++ b/audio/libfreeaptx/libfreeaptx.SlackBuild
@@ -95,11 +95,8 @@ find $PKG -print0 | xargs -0 file | grep -e "executable" -e "shared object" | gr
 mkdir -p $PKG/{install,usr/doc/$PRGNAM-$VERSION}
 
 cp -t $PKG/usr/doc/$PRGNAM-$VERSION \
-    COPYING README && \
-    cp $CWD/$PRGNAM.SlackBuild $PKG/usr/doc/$PRGNAM-$VERSION/
-
-mkdir -p $PKG/install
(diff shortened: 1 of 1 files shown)


willysr:
@thnkman can you fix your name and email so it will be listed in the git commit? It says unknown here and also put category/package as the commit message

audio/libfreeaptx/libfreeaptx.SlackBuild
aclemons, line 1: The permissions here should remain 644

audio/libfreeaptx/libfreeaptx.SlackBuild
```diff
@@ -92,12 +92,11 @@ rm -f $PKG/{,usr/}lib${LIBDIRSUFFIX}/*.la
 find $PKG -print0 | xargs -0 file | grep -e "executable" -e "shared object" | grep ELF \
   | cut -f 1 -d : | xargs strip --strip-unneeded 2> /dev/null || true
 
-mkdir -p $PKG/usr/doc/$PRGNAM-$VERSION
```
aclemons, line 95: I don't think should not arbitrarily deviate from the template.

Things coming from `$CWD` should typically be installed with `cat` or `install` instead of copying them. The items in `$CWD` have of course not had their owner, group, and permissions normalised as the files from source tarball have been.

You can change `cp COPYING README $PKG/usr/doc/$PRGNAM-$VERSION` to `cp -a COPYING README $PKG/usr/doc/$PRGNAM-$VERSION` to preserve these attributes on those files though.

willysr:
@thnkman please do it here as well

PR #4342, opened 2024-01-01
network/dnscrypt-proxy | fixed some permission issues.

thnkman:
Please have a look at my doinst.sh solution and give me some feedback whether or not it is a good solution.


network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
```diff
@@ -22,101 +23,87 @@
...
+    \( -name 'README.Slackware' \) -exec cp {} $PKG/usr/doc/$PRGNAM-$VERSION/ \; -o \
+    \( -name "$PRGNAM.SlackBuild" \) -exec cp {} $PKG/usr/doc/$PRGNAM-$VERSION/ \; -o \
+    \( -name "$PRGNAM" \) -exec cp -a {} $PKG/usr/bin/ \; -o \
+    \( -name "$PRGNAM.toml" \) -exec cp {} $PKG/etc/$PRGNAM/ \; -o \
+    \( -name "rc.$PRGNAM" \) -exec cp {} $PKG/etc/rc.d/ \; -o \
+    \( -name 'slack-desc' \) -exec cp {} $PKG/install/ \;
 
-mkdir -p $PKG/install
-cat $CWD/slack-desc > $PKG/install/slack-desc
 cat $CWD/doinst.sh > $PKG/install/doinst.sh
```
aclemons, line 106: I appreciate your submission and your updating this build, especially since we have no updates from T3slider for ~3 year now, but sorry, please read this page for notes:
https://slackbuilds.org/guidelines/

Especially this part:

> Follow our template scripts as much as possible. Deviations are fine where needed, but don't change things to be "more clever" or some such - we like consistency

We have something like 9000 scripts in the repo and yours is going to be an outlier if we approve it in this way. It is much simpler if everyone uses a similar pattern, even if there are other ways to do it.

network/dnscrypt-proxy/doinst.sh
```diff
@@ -1,27 +1,18 @@
...
+# files to these directories, without having to change default
+# root:root in slackware.
+#
+# This is a workaround, might not be a elegant solution. 
 
-preserve_perms etc/rc.d/rc.dnscrypt-proxy.new
-config etc/default/dnscrypt-proxy.new
-config etc/dnscrypt-proxy/dnscrypt-proxy.toml.new
+PRGNAM=dnscrypt-proxy
+DNSCRYPT_USER=$(cat /usr/doc/dnscrypt-proxy-2.1.5/dnscrypt-proxy.SlackBuild | grep -oP '\${DNSCRYPT_USER:-\K[^}]*')
```
aclemons, line 11: Please don't do this - there are examples of simply running sed on the doinst.sh and piping the output into $PKG/install which is a much better approach.

Here is an example with rc files:

https://github.com/SlackBuildsOrg/slackbuilds/blob/9cbce97d56eec2622da49d6acd43037c77da986e/network/postgrey/postgrey.SlackBuild#L109-L114

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
```diff
@@ -22,101 +23,87 @@
...
+mkdir -p $TMP/$PRGNAM-$VERSION $PKG $OUTPUT
 cd $TMP
 rm -rf $PRGNAM-$VERSION
-mkdir -p $PRGNAM-$VERSION/src/$DOMAIN/$ORG
-cd $PRGNAM-$VERSION/src/$DOMAIN/$ORG
-tar xvf $CWD/$PRGNAM-$VERSION.tar.gz
-mv $PRGNAM-$VERSION $PRGNAM
-cd $TMP/$PRGNAM-$VERSION
+
+if [ $ARCH == i586 ]; then
```
aclemons, line 77: `$ARCH` can be passed in through the env - i.e. `# ARCH=i686 bash dnscrypt-proxy.SlackBuild` so you need to handle that here too.

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
```diff
@@ -22,101 +23,87 @@
...
+    tar xvf $CWD/$PRGNAM-linux_i386-$VERSION.tar.gz --transform="s/linux-i386/$PRGNAM-$VERSION/"
+elif [ $ARCH == x86_64 ]; then
+    tar xvf $CWD/$PRGNAM-linux_x86_64-$VERSION.tar.gz --transform="s/linux-x86_64/$PRGNAM-$VERSION/"
```
aclemons, line 80: Please add an else with `exit 1` and maybe an echo if you do not support any other arches so the scripts fails fast for people.

network/dnscrypt-proxy/rc.dnscrypt-proxy
```diff
@@ -1,134 +1,79 @@
...
 status() {
-    for i in `/usr/bin/seq 0 $((${#DNSCRYPTCONFIG[@]}-1))`
-    do
-        status_instance $i
-    done
+	if [ -z $PID ]; then
+		echo "Not running"
+		exit 1
+	elif kill -0 $PID; then
+		echo $(ps aux | grep $PID)
```
aclemons, line 47: `echo` here is redundant - but do you even need this line at all?

network/dnscrypt-proxy/rc.dnscrypt-proxy
```diff
@@ -1,134 +1,79 @@
...
-    PID=$(cat ${PIDFILE[$1]})
-    if [ -z "$PID" ]; then
-        echo "PID file is empty! dnscrypt-proxy (instance $1) does not appear to be running, but there is a stale PID file."
-    elif kill -0 $PID ; then
-        echo "dnscrypt-proxy (instance $1) is running."
-    else
-        echo "dnscrypt-proxy (instance $1) is not running, but there is a stale PID file."
-    fi
-}
+PRGNAM=dnscrypt-proxy
```
aclemons, line 5: This variable seems unused?

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
```diff
@@ -22,101 +23,87 @@
...
+fi
+
+cd $PRGNAM-$VERSION
 chown -R root:root .
-find -L . \
+
+find -L . $CWD \
```
aclemons, line 86: Are you sure this should be `$CWD`? That is the directory of the slackbuild, sources, slack-desc etc?

aclemons (changes requested):

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
thnkman: > I appreciate your submission and your updating this build, especially since we have no updates from T3slider for ~3 year now, but sorry, please read this page for notes: https://slackbuilds.org/guidelines/
> 
> Especially this part:
> 
> > Follow our template scripts as much as possible. Deviations are fine where needed, but don't change things to be "more clever" or some such - we like consistency
> 
> We have something like 9000 scripts in the repo and yours is going to be an outlier if we approve it in this way. It is much simpler if everyone uses a similar pattern, even if there are other ways to do it.

I will be updating the script to reflect your remarks and have it fall in line with the SBo template. I appreciate your feedback :)

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
thnkman: This wasn't suppose to be there.

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
thnkman: I suppose a simple  .. == i?86 will suffice here?

network/dnscrypt-proxy/rc.dnscrypt-proxy
thnkman: I will be cleaning the rc script up. I forgot to check the sanity of that script before pushing.

thnkman:
Once the script passes the review, i will make a clean PR

willysr:
@sbo-bot: build network/dnscrypt-proxy

network/dnscrypt-proxy/dnscrypt-proxy.SlackBuild
```diff
@@ -97,11 +97,13 @@ mkdir -p $PKG/usr/{bin,doc/$PRGNAM-$VERSION} \
     $PKG/var/{log,run}/$PRGNAM \
     $PKG/install
 
-install -D -m 0644 LICENSE $CWD/README.Slackware $CWD/$PRGNAM.SlackBuild \
-    $PKG/usr/doc/$PRGNAM-$VERSION/
-install -D -m 0644 $PRGNAM $PKG/usr/bin/ 
-install -D -m 0644 $CWD/$PRGNAM.toml $PKG/etc/$PRGNAM/
-install -D -m 0644 $CWD/rc.$PRGNAM $PKG/etc/rc.d/
+cp -a *.txt $PKG/etc/$PRGNAM/
```
willysr, line 100: etc should be used only for configuration files
what is the content of those txt files?

thnkman: They are sample configuration files

willysr:
@thnkman i think you solved all the review, can you send a new clean PR?
this one doesn't apply cleanly on my branch
also, make sure to use "category/package: commit message" as the commit message

PR #4345, opened 2024-01-02
Update to build against Qt5

spillner:
The original SlackBuild linked against Qt4, which is now deprecated in 15.0 and removed from -current.


willysr:
@sbo-bot: build development/xxdiff

willysr:
merged

PR #4350, opened 2024-01-03
system/lxqt-config: Add lxqt-menu-data dependency

isaackwy:
lxqt-menu-data already installs the following files:
```
/etc/xdg/menus/lxqt-config.menu
/usr/share/desktop-directories/lxqt-settings-lxqt.directory
/usr/share/desktop-directories/lxqt-settings-other.directory
/usr/share/desktop-directories/lxqt-settings-system.directory
```

I added a patch that prevents lxqt-config from installing these files.
The patch also explictly sets a requirement for lxqt-menu-data.
This is similar to the patch I added for lxqt-panel.

That being said, I have not added an explicit lxqt-menu-data requirement to the lxqt-panel patch (the dependency is only listed in lxqt-panel.info). I think I will send a PR for making those modifications next week.


willysr:
merged

PR #4367, opened 2024-01-03
Fix email addresses

isaackwy:
I just discovered an email address typo within a few of my recently uploaded packages.

The email address should be isaacyu@protonmail.com, not isaacyu1@isaacyu1.com.
I no longer use isaacyu1@isaacyu1.com.


willysr:
merged

PR #4406, opened 2024-01-06
desktop/lxqt-panel: Explicitly require lxqt-menu-data for compliation

isaackwy:
Last week, I was especially concerned that I only changed the dependency info (from lxmenu-data to lxqt-menu-data) without adding a patch preventing the installation of files already covered by lxqt-menu-data. This risked breaking the lxqt-panel ui (especially with lxmenu-data uninstalled).

I made the correction accordingly - however, (unlike with lxqt-panel) I did not explicitly set an lxqt-menu-data requirement for compilation.

This pull request adds this requirement.

I decided not to bump the build number (I'm not interested in getting other users to recompile lxqt-panel).


willysr:
merged

PR #4414, opened 2024-01-06
python/python3-webcolors: Update for 1.13

isaackwy:
I previously added python3-webcolors as a dependency for jupyter-server.
I'll keep maintaining it for now - if there is no more reason to maintain python3-webcolors, I will orphan it along with previous jupyter_server dependencies (ex. python3-uri_template).


willysr:
@sbo-bot: build python/python3-webcolors

willysr:
merged

PR #4416, opened 2024-01-06
python/python3-anyio: Update README

isaackwy:
fourtysixandtwo's python3-setuptools-opt and python3-setuptools-scm-opt can be used to update python3-anyio.
I have decided to remove the blurb that states "Newer versions would remove a newer python-setuptools..."

However, I am not updating python3-anyio (even the most recent versions of jupyter_server work on any python3-anyio >= 3.1.0).


willysr:
merged

PR #4418, opened 2024-01-06
development/jupyter_packaging: Update README

isaackwy:
Fourtysixandtwo's python3-setuptools-opt package has allowed the building of packages with setuptools >= 60.

For example, jupyter_packaging 0.12.3, which requires setuptools >= 60.2.0, may be able to be built with the additional python3-setuptools-opt dependency.

However, I suspect that setuptools is a runtime dependency for jupyter_packaging (rather than merely being a build-time one).

That being said, I will not update jupyter_packaging (I might update it when Pat updates Slackware to 15.1, though).

### DIFF
--- a/development/jupyter_packaging/README
+++ b/development/jupyter_packaging/README
@@ -1,5 +1,2 @@
 Jupyter Packaging contains tools to help build and install Jupyter
 Python packages.
-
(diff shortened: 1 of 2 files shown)


willysr:
merged

PR #4421, opened 2024-01-06
python/python3-pywayland: Update README

isaackwy:
fourtysixandtwo's python3-setuptools-opt allows me to compile later versions of python3-pywayland.

However, I still will not update python3-pywayland.
If the pywayland changelog (https://github.com/flacjacket/pywayland/releases) is of any indication, pywayland > 0.4.14 requires a higher version of wayland-protocols (1.31 for pywayland 0.4.15 vs 1.25 for Slackware 15.0).


willysr:
merged

PR #4455, opened 2024-01-09
Vine

Ythogtha:
Added python3-vine through submit form.
This removes vine and updates python3-kombu to the new name.

I've taken over vine since :
 * Previous maintainer never answered.
 * vine is only required by python3-kombu which is part of requirements for Kallithea, through python3-celery
 * I maintain Kallithea, python3-celery and python3-kombu


willysr:
merged

PR #4460, opened 2024-01-09
desktop/rss-guard: updated for version 4.6.3

Ythogtha:
That one is a bit tricky.
rssguard builds now by default for qt6, but it can still build for qt5.
So there is an autodetection in the script, to automatically fallback on qt5 when qt6 isn't there, which is the default.
I also added some build options available, mostly the possibility to use mpv instead of ffmpeg for media playback.
That's a new functionality of rss-guard.

But I do have trouble with the qt6 version.
And because it seems the problem is on my end (wrong path being returned by qt6 for itself, for whatever reason), I'm guessing its ok for most people.

With the right ENV variable set, I can use the qt6 version, and it works fine.
There's no telling the difference between the two versions, visually, so that's really up to people, Qt5 version is perfectly fine.


willysr:
@sbo-bot: build desktop/rss-guard

willysr:
merged

PR #4467, opened 2024-01-10
libraries/allegro: updated for version 5.2.9.0

Ythogtha:
With fixes on lib/lib64.

Apparently allegro installs libs onto both /usr/lib/ and /usr/lib64/ when asked for lib64.
Also it creates /usr/lib/cmake but /usr/lib64/pkgconfig.
Everything looks ok in 32bits.
So I made changes, removing duplicate libs on /usr/lib/ when LIBDIRSUFFIX exists, and making sure cmake/ and pkgconfig/ directories are all on /usr/lib$LIBDIRSUFFIX/

I tested on x86_64, with 32bits configuration on x86_64 (and it generates an i686 proper package, with 32bits libs), and on Aarch64 (which has the same problem as x86_64), unfortunately my ARM box is too old and needs a good reinstallation, but I don't have enough time for that.

Also, B-em, opensurge and mininim are working fine, so I guess that works.


willysr:
@sbo-bot: build libraries/allegro

willysr:
merged

PR #4505, opened 2024-01-11
audio/yabridge: Fix buildtype and let every exit have a unique number

noisecode3:
Only take the last commit this time.

### DIFF
--- a/audio/yabridge/yabridge.SlackBuild
+++ b/audio/yabridge/yabridge.SlackBuild
@@ -31,7 +31,7 @@ TAG=${TAG:-_SBo}
 PKGTYPE=${PKGTYPE:-tgz}
 ARG_CLAP=${ARG_CLAP:-NO}
 ARG_VST3=${ARG_VST3:-YES}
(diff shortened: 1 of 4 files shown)


audio/yabridge/yabridge.SlackBuild
```diff
@@ -22,7 +22,7 @@
 #  OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 #  ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-cd $(dirname $0) || exit 1 ; CWD=$(pwd)
+cd $(dirname $0) || exit 69 ; CWD=$(pwd)
```
willysr, line 25: i don't think this is correct
if you want complete list, it's here: https://www.cyberciti.biz/faq/linux-bash-exit-status-set-exit-statusin-bash/
but i think simply exit 1 is just fine

willysr:
also, if you need to rebase latest master or else this PR will include 3 commits

PR #4506, opened 2024-01-11
system/greetd: Updated mirror link in greetd.info

mac-a-r0ni:
update info to use github mirror, as main site is down for days now.


willysr:
merged

PR #4537, opened 2024-01-13
libraries/grpc: Remove Python 3 support

isaackwy:
I plan on uploading python3-grpcio this week.
At the moment, the grpc SlackBuild builds Python 3 files.

Due to the overlap between grpc and python3-grpcio files, I would like Python 3 support removed from grpc (and then upload python3-grpcio as a separate SlackBuild).

bear and sysdig (dependencies of python3-grpcio) do not require Python 3 support for grpc, so they remain unaffected by this change.

I would like to do more with this SlackBuild (for example, add abseil-cpp to .info as a DEP, rather than downloading abseil-cpp and building with that). I would also like to get grpc updated to a later version.
However, I do not know enough about grpc at the moment to make these changes.


willysr:
merged

PR #4558, opened 2024-01-15
libraries/grpc: Bump build number

isaackwy:
Sorry, I forgot to bump grpc's build number.

Users might forget to update the grpc SlackBuild (with Python 3 support removed) and then install python3-grpcio.
python3-grpcio will then overwrite the Python 3 files already installed by grpc (grpc should not install any Python 3 files before the user installs python3-grpcio).


willysr:
there's no need to bump for trivial changes

PR #4560, opened 2024-01-15
python/python3-lsp-black: Update README

isaackwy:
python3-lsp-black 2.0.0 (the newest version) requires python3-black >= 23.11.0 (for context, python3-black is held back at version 22.12.0).


willysr:
merged

PR #4563, opened 2024-01-15
python/python3-lsp-jsonrpc: Update for 1.1.2

isaackwy:
This update is not necessary for the current version of python3-lsp-server at SlackBuilds.org (1.6.0).
However, this will become necessary when updating python3-lsp-server to later versions (i.e. at >=1.8.0).

In addition, given that the python3-lsp-jsonrpc SlackBuild has now added a python3-setuptools-opt dependency, the python3-lsp-server SlackBuild should have its python3-setuptools-opt dependency removed.


willysr:
merged

PR #4578, opened 2024-01-17
python/python3-kodipydent: Fix pip check warning

bassmadrigal:
Update dependency name to newer alternate and remove version requirement so `pip check` doesn't complain anymore.


willysr:
merged

PR #4619, opened 2024-01-18
Aarch64

Ythogtha:
Adding the aarch64 SLCKFLAGS and LIBDIRSUFFIX section.
Sometimes fixing docs, removing INSTALL instructions, when sbopkglint complains.
Also, and only for SDL2_sound and libjodycode, removes lib/*.a, which are unneeded by other packages requiring those ones (namely ags and jdupes only).
They've been kept for dumb, as they are used.
Also fixed an old man page for sshuttle, that disappeared some time ago but was still generating a gzip of an empty file...
And the .desktop file of focuswriter-qt6 now checks with sbopkglint also.

Nothing here should warrant a change in build numbers.

Hoping that this quite huge PR will be ok, and not too much work with the CI. It's quite easy to review by eye though.


willysr:
merged

PR #4711, opened 2024-01-24
desktop/dwmblocks: Update for 20231219 (+switch to UtkarshVerma's implementation)

isaackwy:
I am not using the torrinfail build of dwmblocks. I am instead using UtkarshVerma's build (which runs asynchronously and handles mouse clicks).

In effect, I am "removing" the SlackBuild for torrinfail's dwmblocks and "uploading" the SlackBuild for UtkarshVerma's.

UtkarshVerma's build of dwmblocks has an issue though - if do the following: 
- Displaying volume from a pulseaudio sound server 
- Wipe cache with Bleachbit after using a web browser (ex. firefox, chromium-ungoogled)

The window manager (dwm) then faces the potential of crashing.

In the README, I added suggestions for getting around this issue. The cause of the problem is pulseaudio (hence, I am suggesting users to manually upgrade pipewire to a newer version and to use a version of wireplumber newer than the one available at SBo).


willysr:
@sbo-bot: build desktop/dwmblocks

willysr:
merged

PR #4723, opened 2024-01-24
graphics/vimiv-qt: Update vimiv-qt.SlackBuild

isaackwy:
The SlackBuild should have used make install rather than python3 setup.py install.

python3 setup.py install alone does not install man files, desktkop entry, etc.


willysr:
merged

PR #4875, opened 2024-02-06
graphics/xbmbrowser: The URL update.

bocke:
The recent homepage doesn't mention xbmbrowser, so I am using the archived version of url from IA that does.


willysr:
merged

PR #4944, opened 2024-02-09
Add note on why python3-pandas and ipython cannot be updated

bassmadrigal:
python/python3-pandas: Note on why we can't update
python/ipython: Note on why we can't update


aclemons (approved):
LGTM

PR #5062, opened 2024-02-18
Add libei

nater1983:


aclemons:
Thanks for the submission @nater1983 , however new submissions need to first go through the submission form on slackbuilds.org. This populates the db, keywords etc. After that, you can submit updates here via pull requests.

See notes here:

https://slackbuilds.org/SBo-Github-Workflows.txt

PR #5083, opened 2024-02-21
Lxml/cssselect python2/python3 split

bassmadrigal:
Simply split the lxml and cssselect dependencies to their python2 and/or python3 couterparts to aid in the removal of lxml and cssselect combined package.


willysr:
merged

PR #5181, opened 2024-02-24
libraries/libkml: fixed outdated libkml build

wsdookadr:
Prior to this PR, the libkml package was not building correctly.
It was trying to pull a tarball that has been marked vulnerable [1] and because the url for it was renamed, it wasn't able to fetch it.

The fix was to patch the CMake file for libkml to use a newer version of uriparser.
Tested locally on a Slackware 15 instance and everything went well.

Please advise regarding the version bump if one is required.
Maybe we should use 1.3.0-1, not sure.

[1] https://sourceforge.net/projects/uriparser/files/Sources/0.7.5/


wsdookadr:
@sbo-bot: build libraries/libkml

willysr:
@sbo-bot: build libraries/libkml

willysr:
merged

PR #5196, opened 2024-02-25
Update afdko.info

lecramyajiv:
checkoutlinesufo tool in afdko needs 
python3-ufonormalizer, python3-fontParts and 
Python3-fontPens for it to work.


willysr:
you need to ask maintainer instead of directly made any PR

PR #5218, opened 2024-02-26
desktop/wofi: Updated for version 1.4.1.

perrin4869:


aclemons:
@sbo-bot: build desktop/wofi

perrin4869:
hm... did it somehow fail to download the source files?

desktop/wofi/wofi.info
```diff
@@ -1,8 +1,8 @@
 PRGNAM="wofi"
-VERSION="1.3"
+VERSION="1.4.1"
 HOMEPAGE="https://hg.sr.ht/~scoopta/wofi"
-DOWNLOAD="https://hg.sr.ht/~scoopta/wofi/archive/v1.3/wofi-1.3.tar.gz"
-MD5SUM="a185b368acebb44f221d0f8910e971f6"
+DOWNLOAD="https://hg.sr.ht/~scoopta/wofi/archive/v1.4.1.tar.gz"
```
aclemons, line 4: It's the format used for the url. Can you update to use: https://hg.sr.ht/~scoopta/wofi/archive/v1.4.1/wofi-v1.4.1.tar.gz

Then it will work with browsers or clients respecting the `content-disposition` header and also if you just use wget and it takes the last path as the filename. Both will result in the same filename.

perrin4869: yeah, that was what I tried at first actually, but neither the browser nor `wget` manage to fetch via `https://hg.sr.ht/~scoopta/wofi/archive/v1.4.1/wofi-v1.4.1.tar.gz`, which is why I reverted to `v1.4.1`...
Actually, the old path, `https://hg.sr.ht/~scoopta/wofi/archive/v1.3/wofi-1.3.tar.gz` stopped working too, so it might be a bug in sourcehut, I'll confirm with support

perrin4869: yeah, I just got back from `sourcehut` support:

```
I think you're mistaken? I don't think that this
URL was ever supported. That looks more like the format GitHub uses.
```

Unfortunately we can't specify a filename as a last path parameter in the case of sourcehut

aclemons: Ok I see. Could you handle both cases for the filename in the script then please? If someone downloads with a browser (clicking the links on slackbuilds.org), it will result in the full filename. wget without `--content-disposition` will just be v1.4.1.tar.gz. Thanks

willysr:
@sbo-bot: build desktop/wofi

willysr:
merged

desktop/wofi/wofi.info
aclemons: Thanks

perrin4869: actually, I noticed the script already supported both filenames, but the server had changed their content disposition to include the `v` in the filename

PR #5255, opened 2024-02-29
Marcel/winff

lecramyajiv:
I have updated the info file so that it will now   
download properly. I checked it with sbo-create tool
it downloads properly and also builds without any problem


willysr:
this pr is almost identical with the previous one
we only keep 1st level dependency, not the whole dependency list
cffsubr list adfko and all the new deps listed in adfko changes as well

lecramyajiv:
this is meant for winff only not for cffsubr I must have done something wrong with this

PR #5270, opened 2024-02-29
python/pyinotify: New maintainer, remove Python 2 support

isaackwy:
I don't think Larry Hajali is maintaining SlackBuilds anymore.

I would like to take over maintenance of pyinotify (a dependency of OpenSnitch).

Also, I am removing Python 2 support from pyinotify (both watch-fs and OpenSnitch build with python3 rather than python2).


willysr:
@sbo-bot: build python/pyinotify

willysr:
merged

PR #5319, opened 2024-03-02
perl/perl-html-parser: Updated for version 3.82.

pyllyukko:


willysr:
@pyllyukko HTTP::Headers is not installed

pyllyukko:
> @pyllyukko HTTP::Headers is not installed

I'll fix that during the weekend and force push here in the same PR.

willysr:
if you can push the changes to your branch, i can recheck it quickly for you, no need to check later

pyllyukko:
So `HTTP::Headers` doesn't exist currently in SBo, but is provided by [perl-http-message](https://slackbuilds.org/repository/15.0/perl/perl-http-message/). Should I create a new `SlackBuild` for `HTTP::Headers` or use the `perl-http-message` package as a dependency? `perl-http-message` probably brings bunch of other functionality that is not absolutely required by `perl-html-parser`. Having both `HTTP::Headers` and `HTTP::Message` probably also has overlapping files in the packages, which would not be optimal.

willysr:
you can try to use perl-http-message first

willysr:
@sbo-bot: build perl/perl-html-parser

willysr:
seems like we can't use perl-http-message as it will cause circular dependency

pyllyukko:
> seems like we can't use perl-http-message as it will cause circular dependency

I started looking at packaging [HTTP::Headers](https://metacpan.org/pod/HTTP::Headers), but the download link is for the [exact same HTTP::Message](https://cpan.metacpan.org/authors/id/O/OA/OALDERS/HTTP-Message-6.45.tar.gz) :D So apparently you can't even have `HTTP::Headers` separately without `HTTP::Message`. What now?

willysr:
have you checked the upstream project requirement?

pyllyukko:
> have you checked the upstream project requirement?

I did and they should be good, but once I checked the deps for [HTTP::Message](https://metacpan.org/pod/HTTP::Message), I noticed that it doesn't actually require this package. So we need to update `HTTP::Message` and remove this as a dependency and then it's not circular anymore.

I'll wait for #5324 to land into upstream, update the deps on `HTTP::Message`, create a PR for that and after the `HTTP::Message` fix is merged this should be good to go :)

pyllyukko:
After merging #5551, this should work.

willysr:
@sbo-bot: build perl/perl-html-parser

willysr:
merged

PR #5336, opened 2024-03-02
cffsubr resubmit

lecramyajiv:
removed the updated deps for the
afdko in the info file. we are using
the locally installed version of afdko instead of downloading it during install


willysr:
@sbo-bot: build python/python3-cffsubr

willysr:
@lecramyajiv please never change the permission to 755 for the SlackBuild
keep it 644

PR #5338, opened 2024-03-02
python/lxml: Removed (split into python2-lxml and python3-lxml)

isaackwy:
Thanks to @bassmadrigal for splitting lxml into python2-lxml and python3-lxml.

One more thing though - as a final step, I would like lxml removed from SlackBuilds.org.
lxml is now a standalone SlackBuild - it is no longer a dependency for anything.


bassmadrigal:
Thanks! I forgot I was going to submit that request for the last public update and forgot.

isaackwy:
You're welcome Jeremy!

willysr:
merged

PR #5343, opened 2024-03-03
python/cssselect: Removed (split into python2-cssselect and python3-cssselect

bassmadrigal:
Thanks to @isaackwy reminding me, I also split cssselect and can have the original removed.


willysr:
merged

PR #5345, opened 2024-03-03
winff slackbuild update

lecramyajiv:
Removed unnecessary functions in doinst. lazbuild will no longer create object code in /usr/share/lazarus since it will now build from /tmp/ and use it as build dir.


willysr:
@sbo-bot: build multimedia/winff

multimedia/winff/winff.SlackBuild
```diff
@@ -70,12 +69,12 @@ set -e
 trap 'echo "$0 FAILED at line $LINENO!" | tee -a  $OUTPUT/error-${PRGNAM}.log' ERR
 
 rm -rf $PKG
+rm -rf $TMP
```
willysr, line 72: you can't rm $TMP as it will break the machine

willysr: rm $PKG already delete the directory created by extracting the source, no need to delete the whole $TMP

PR #5351, opened 2024-03-04
system/lxqt-config: Remove lxqt-themes dependency

isaackwy:
I built lxqt-config without lxqt-themes successfully.

lxqt-config does not require lxqt-themes (as a build or runtime dependency) - however, lxqt-themes is useful for the LXQt desktop environment.


willysr:
merged

PR #5357, opened 2024-03-06
corrected the slackbuild

lecramyajiv:
removed the line rm -rf   $TMP


willysr:
@sbo-bot: build multimedia/winff

lecramyajiv:
@willysr  I haven't changed the winff.info file if you see the pull request [5284](https://github.com/SlackBuildsOrg/slackbuilds/pull/5284) that downloaded the files and executed it.which is the same info file. I could download it using sbo-create and the slackbuild found it with no problems in my VM. This is my info file 

`<`
PRGNAM="winff"
VERSION="1.6.3"
HOMEPAGE="https://github.com/WinFF/winff"
DOWNLOAD="https://github.com/WinFF/winff/archive/refs/tags/winff-1.6.3/winff-1.6.3.tar.gz"
MD5SUM="7a85f02bd23acef5b916166f83586e82"
DOWNLOAD_x86_64=""
MD5SUM_x86_64=""
REQUIRES="libqtpas lazarus"
MAINTAINER="Vijay Marcel"
EMAIL="vijaymarcel@outlook.com"

and this is the error causing para in the winff slackbuild 
`<`
rm -rf $PKG
mkdir -p $TMP $PKG $OUTPUT
cd $TMP
rm -rf $PRGNAM-$VERSION
tar xvf $CWD/$PRGNAM-$VERSION.tar.gz
cd $PRGNAM-$PRGNAM-$VERSION

At this point I give up, I don't know why it causes this problem because it works with wget,sbo-create, even browser downlaoded it. By any chance did the downloaded file had the file name winff-winff-1.6.3.tar.gz I couldn't see the log so
I don't know,any help would be appreciated. 
There is another way i could try using the specific commit to download that version of that archive

multimedia/winff/winff.SlackBuild
```diff
@@ -73,9 +72,8 @@ rm -rf $PKG
 mkdir -p $TMP $PKG $OUTPUT
 cd $TMP
 rm -rf $PRGNAM-$VERSION
-rm -rf $PRGNAM-$LONGNAM
-tar xvf $CWD/$PRGNAM-$LONGNAM.tar.gz
-cd $PRGNAM-$LONGNAM
+tar xvf $CWD/$PRGNAM-$VERSION.tar.gz
+cd $PRGNAM-$PRGNAM-$VERSION
```
aclemons, line 76: You can see the issue here. For github downloads, if you have a different filename to the source directory the basename of the file used in the tar line, then the URL used is likely to give different results with wget and a browser etc.

When I follow the link (here in a browser) to the source from the info file in the PR, it results in the name you use here in `cd` namely, winff-winff-1.6.3.tar.gz. That's because the name of the tag is `winff-1.6.3` and then the project name is prefixed:

winff

The fix is to use this full name for all three things: cd, tar, and rm.

Then url in the info file should be:

https://github.com/WinFF/winff/archive/winff-1.6.3/winff-winff-1.6.3.tar.gz

That will give the same filename with wget, a browser, wget with `--content-disposition` etc.

Dave wrote a document with more details here: https://slackbuilds.org/GITHUB_URLs.txt

PR #5476, opened 2024-03-12
multimedia/spot: Updated for version 0.4.1

perrin4869:


willysr:
@sbo-bot: build multimedia/spot

willysr:
@sbo-bot: build multimedia/spot

willysr:
i don't think this can be compiled under Slackware 15.0 as it requires newer glib2

perrin4869:
Yeah I was fearing it might be the case, guess I'll close this for now.
Just out of curiosity, do you know why rust16 is failing to find the shared lib?

PR #5509, opened 2024-03-14
Kallithea

Ythogtha:
Sorry for two updates on the same week, I didn't think I could fix for SQLAlchemy 2 this week.
So this second commit adds two patches, on for a bug in bleach usage preventing display of README on repositories.
And the other (quite big) one for compatibility with SQLAlchemy 2.

As far as I have been able to test, and look inside Kallithea's sources, this should work on both current SQLAlchemy (1.4.52) and on the next one by 46&2 (2.0.29).

I also removed a lot of warning (all of them really) on startup.
Next step will be to send a pull request upstream, but that shouldn't concern us here.


willysr:
@sbo-bot: build python/Kallithea

willysr:
merged

willysr:
@Ythogtha since SQLAchemy 2 won't be merged this week, is it ok for me to merge the last commit?

Ythogtha:
> @Ythogtha since SQLAchemy 2 won't be merged this week, is it ok for me to merge the last commit?

@willysr Yes, it's ok, that one works with both SQLAlchemy 1.4 and 2.0, so it can be merged anytime, it'll work.

willysr:
merged

PR #5520, opened 2024-03-15
python/hg-git: reverted to version 1.0.5, because mercurial is too old

Ythogtha:
Looks like I was a little trigger-happy with the last update, it is incompatible with Slackware's 15.0 mercurial version.
Reverting to last compatible version.


willysr:
merged

PR #5523, opened 2024-03-15
Buildbot

Ythogtha:
To allow for updating SQLAlchemy to version 2.x, buildbot needs a legacy version.
This is it: requiring the approbation of the newly submitted SQLAlchemy-legacy package.

To be noted : SQLAlchemy was a requirement of buildbot through python3-alembic, but is really used directly by buildbot.
python3-alembic can work with both SQLAlchemy versions.
But in the requirement tree, there'll be that :
buildbot
 |- python3-alembic
 |\-SQLAlchemy
 \- SQLAlchemy-legacy

And both SQLAlchemy and SQLAlchemy-legacy are incompatible with each other.
People will have to read the READMEs.

### DIFF
--- a/python/buildbot/README
+++ b/python/buildbot/README
@@ -12,3 +12,7 @@ tracked over time, are more visible, and are therefore easier to
 improve.
 
 See also buildbot-worker (the worker component of BuildBot).
+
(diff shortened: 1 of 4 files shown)


willysr:
merged

PR #5540, opened 2024-03-16
Moving podcast parser to a better/correct category.

DinizBortolotto:
Moving podcast parser to the correct category since, as stated in README, it is not a multimedia application.
It's a library from the gPodder to parsing RSS- and Atom-based podcast feeds.


willysr:
merged

PR #5551, opened 2024-03-16
perl/perl-http-message: Updated dependencies.

pyllyukko:
Bumped the revision number, even though the package itself is not changed in any way, but only the metadata in `.info`.


willysr:
merged

PR #5558, opened 2024-03-16
desktop/xsecurelock: Update build - make pandoc-bin an optional dependency

isaackwy:
pandoc-bin (or pandoc) is only required at build-time for building man pages.
I have now made pandoc-bin an optional dependency.

This SlackBuild now autodetects whether pandoc is installed:
`pandoc --version >/dev/null 2&>1`

I also have made additional changes to the README (ex. description of mpv as an optional requirement for the saver_mpv module).


willysr:
merged

PR #5564, opened 2024-03-17
misc/bitwarden-desktop

maravtdm:
Rollback to  version 2024.1.0
version 2024.2.1 seems to have been compiled with a higher version of glibc


willysr:
@maravtdm this is not based on latest master branch

PR #5586, opened 2024-03-19
misc/bitwarden-desktop: revert to version (2024.1.0) that works

maravtdm:
Revert to 2024.1.0 since 2024.2.1 was compiled with newer glibc


willysr:
merged

PR #5642, opened 2024-03-23
python/pyinotify: Removed (renamed to python3-pyinotify)

isaackwy:
I now have replace pyinotify with python3-pyinotify.
Please remove pyinotify from SlackBuilds.org.


willysr:
merged

PR #5645, opened 2024-03-23
system/jq: Fix build test failures associated with non-English modules

isaackwy:
I took a look at the SlackBuilds mailing list.
Thanks to Tonus for mentioning build failures when setting a non-english LANG (ex. fr_FR.UTF-8).
Also, thanks to nnb (nomnombtc) for linking to https://github.com/jqlang/jq/issues/3038.


willysr:
merged

PR #5660, opened 2024-03-24
python/python3-keyring: Update for 25.0.0

isaackwy:
python3-jaraco.context and python3-jaraco.functools are now additional dependencies.
Also, remove python3-setuptools-opt from the list of dependencies (as python3-jaraco.classes itself depends on python3-setuptools-opt).


willysr:
@sbo-bot: build python/python3-keyring

PR #5692, opened 2024-03-29
libmediainfo & mediainfo version bumps to 24.03

bassmadrigal:


aclemons:
@sbo-bot: build libraries/libmediainfo

aclemons:
@sbo-bot: build multimedia/mediainfo

bassmadrigal:
@aclemons, these failures look to be due to the usage of wxPython as the optional dependency to enable the GUI, but testing has only been accomplished with wxGTK3 and is the only one mentioned in the README. Can this be adjusted with the build bot's hintfile?

I'm not very familiar with the wx options, but maybe it needs at least wxPython3 (which already has a dependency of wxGTK3, so it's probably not worth trying).

bassmadrigal:
Looks like the breakage is [due to a change](https://github.com/MediaArea/MediaInfo/commit/e0abab51371fb687b6ac5ad90bc4a25f92889c29) on Mediainfo's side adding a reference to a header not present in wxPython but is present in wxGTK3. I'm not sure if simply reverting this would fix it, but it might continue to get worse if wxPython-2.x is no longer being updated and mediainfo expects things from newer wxwidget programs.

aclemons:
> Can this be adjusted with the build bot's hintfile?

Yes, I will update the hints and retrigger. Thanks for the ping.

aclemons:
@sbo-bot: build multimedia/mediainfo

aclemons (approved):
LGTM

PR #5707, opened 2024-03-30
development/abseil-cpp: Update for 20240116

isaackwy:
@willysr Judging from Arch Linux's [build](https://gitlab.archlinux.org/archlinux/packaging/packages/abseil-cpp) of abseil-cpp:

Arch Linux contains an additional patch named [scoped-mock-log.patch](https://gitlab.archlinux.org/archlinux/packaging/packages/abseil-cpp/-/blob/main/scoped-mock-log.patch)

We're not building tests for either abseil-cpp or protobuf3, so we don't need to disable the "TESTONLY" option inside of the scoped_mock_log library (which is what the patch is doing).

I think abseil-cpp's default behaviour is to not require googletest (SlackBuilds names this gtest) by default.
If the patch were included, then gtest will become an additional requirement for abseil-cpp.

See this comment for further details: https://github.com/abseil/abseil-cpp/issues/1407#issuecomment-1451149066


willysr:
@sbo-bot: build development/abseil-cpp

willysr:
i don't think Patrick also included gtest in Slackware so i guess we can use the same route

willysr:
merged

PR #5713, opened 2024-03-31
development/jupyterlab: Add .desktop file

isaackwy:
jupyterlab 4 has a .desktop file installed by default.

I think it would be a good idea for jupyterlab 3 to have this as well.

Anyway, I am not updating jupyterlab until Slackware 15.1 (or whatever comes after 15.0) gets released.
Any later version of jupyterlab requires python-requests > 2.26.


willysr:
merged

PR #5779, opened 2024-04-03
academic/python3-scipy: Note on why we can't update

bassmadrigal:


bassmadrigal:
I just realized, with 9dd796b, should I increase the build number? The change doesn't affect any building on 15.0 and was previously broken on -current, so it wouldn't have built on there.

willysr:
merged

PR #5781, opened 2024-04-03
python/python3-validators: Version bump to 0.27.0

bassmadrigal:
They've been busy this week...


willysr:
@sbo-bot: build python/python3-validators

willysr:
merged

bassmadrigal:
You can remove it. There was another update tonight (that's like 4 or 5 versions within a couple of days). I'll push out a new update tomorrow with the PYTHONPATH change and new version (which might be before or after the public release, depending on when I get home from work and how soon Willy pushes the update).

willysr:
i will probably push public update in 24h because I have to travel and I won't be able to push the update in normal schedule

PR #5782, opened 2024-04-03
python/python3-platformdirs: Fix building on -current

bassmadrigal:
Thanks to @fourtysixand2 for the nudge.


willysr:
@sbo-bot: build python/python3-platformdirs

willysr:
current moves to 3.11
this will fail to build for sure

bassmadrigal:
Why would this fail on -current? The 3.9 reference is only to grab the setuptools-scm in /opt/ on 15.0. It should still be able to reference the tools in the normal location, right?

willysr:
You need to use dynamic python version detection instead of hard-coding it as 3.9 as current move python3 to python3 3.11

willysr:
i fixed it and merged it

PR #5832, opened 2024-04-08
Comment with reverse dependencies on PRs.

aclemons:
This should give a quick overview of what other scripts could be
affected by the change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,3 +105,52 @@ jobs:
         with:
           issue-number: ${{ github.event.number }}
           body-path: comment-output
+
+  dependencies:
+    name: Compute reverse dependencies
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      pull-requests: write
+    needs: [changes]
+    strategy:
+      matrix:
+        include: ${{ fromJSON(needs.changes.outputs.matrix) }}
+    steps:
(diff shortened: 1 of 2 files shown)


willysr:
How does this works?

aclemons:
Urchlay added support for `text/plain` responses to the reverse dependency lookup at https://slackbuilds.org/advsearch.php. This change will check the updated scripts in a PR/MR and then post a comment showing what the reverse dependencies are. It should help us be aware of scripts that have many dependees or that have important dependees.

For example, there are still scripts in the repo broken by the last `fmt` upgrade. The next time a PR for it is opened, this should comment with:


#### reverse dependencies - libraries/fmt

```
games/0ad
development/Bear
games/RMG
libraries/SDL_audiolib
games/SLADE
misc/fcitx5
libraries/folly
audio/mpd
gis/osgEarth
office/paps
audio/patchage
games/pcsx2
gis/therion
graphics/ueberzugpp
desktop/waybar
```

Next step is to add a new command to the bot to tell it to build all of these so we don't have to send an `@sbo-bot: build xyz` for each.

bassmadrigal:
Just for the simple looks of it, any chance at sorting the output?

aclemons:
> Love this idea! Just for the simple looks of it, any chance at sorting the output?

Good idea, thanks @bassmadrigal. Done now.

willysr:
merged

PR #5944, opened 2024-04-14
libraries/mygpoclient: Fix DOWNLOAD link on info file and SlackBuild comments

DinizBortolotto:
libraries/mygpoclient: Fix DOWNLOAD link on info file and SlackBuild comments


willysr:
merged

PR #5960, opened 2024-04-15
Update xdg-utils-cxx.SlackBuild

lecramyajiv:
Installing test binaries need to be removed in slackbuild if we are to  turn tests off, otherwise it will cause build problems.


willysr:
i have this changes already on my branch
https://git.slackbuilds.org/slackbuilds/commit/?id=4fad22358
https://git.slackbuilds.org/slackbuilds/commit/?id=822c1731f

PR #5973, opened 2024-04-16
multimedia/stremio: Updated to 4.4.168 and audio/bitwig-studio: Updated to 5.1.7

noisecode3:


willysr:
@noisecode3 can you fix the warning/errors from sbolint?

willysr:
@sbo-bot: build multimedia/stremio

noisecode3:
I use virtualbox with slackware 15.0 to see if the program works but I have used sbopkglint on my slackware current and I  did not get this ```error: value "1.5" for key "Version" in group "Desktop Entry" is not a known version``` so I have been doing this wrong apparently... so its just a sed command that change 1.5 to 1.0 in the BitwigStudio.desktop is it really a problem ?

willysr:
you can ignore that :)

willysr:
merged

PR #6121, opened 2024-04-25
system/FreeFileSync: Updated info file, download location.

antonioleal:
system/FreeFileSync: Updated info file, download location.

As per message below:
----------------------------------

Hi all

Andrew has compiled a list of FTBFS (Fail to Build From Source) which are listed here: https://github.com/SlackBuildsOrg/slackbuilds/issues/5827.

If you are the maintainer of those packages, please check them and fix it. If you are using those scripts and have some time, please help us to fix those scripts.

Some of those are just md5sum mismatches, so it's easy to fix, but some are broken due to updates from one of it's deps or maybe there's a chance in upstream but the maintainer didn't notice or maybe even worse, maintainer is no longer active and scripts are abandoned.

You can submit fixes via pull requests, so it's easy to verify using our CI engine.

Thanks

-- 
Willy Sudiarto Raharjo

### DIFF
--- a/system/FreeFileSync/FreeFileSync.info
+++ b/system/FreeFileSync/FreeFileSync.info
@@ -1,7 +1,7 @@
 PRGNAM="FreeFileSync"
 VERSION="13.5"
 HOMEPAGE="https://freefilesync.org/"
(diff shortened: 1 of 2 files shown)


willysr:
merged

PR #6162, opened 2024-04-27
libraries/folly: Updated for version 20240420_c2cf035

earies:


willysr:
@sbo-bot: build libraries/folly

willysr:
@earies you removed fmt which is still needed by this project

earies:
> @earies you removed fmt which is still needed by this project

Added back - was removed some time ago when fmt became part of upstream `l/fmt`

For slackbuild dependencies, when items move from SBo to become officially part of base slackware, it is unclear if these are necessary to be called out (appears so here) as have also noticed a number of others over the years not calling this out if the base pipeline assumes to contain such deps (or this could be a matter of -current assumptions only)

willysr:
@sbo-bot: build libraries/folly

willysr:
@earies still failed to build here

willysr:
@earies any update on this? or can we drop this?

earies:
> @earies any update on this? or can we drop this?

Can you confirm if the build environment is on `-current` with liburing 2.6 (from 5/1/2024 updates)?

willysr:
we use -stable for this VM since SBo is for stable

earies:
> we use -stable for this VM since SBo is for stable

ok so then this likely aligns to the following liburing version

https://mirrors.slackware.com/slackware/slackware64-15.0/source/l/liburing/

which is on 2.1 (9/10/2021) and incompatible - `folly` publishes are weekly and have a number of very recent dependencies each week so it's likely this PR needs to be dropped and any folly slackbuild support can really only be against `-current` (and even current will often not be able to support w/o constant alignment of core dependency updates)

willysr:
then i would suggest to drop back to the version which supports older version of liburing and make a NOTE in the README about that situation.

earies:
> then i would suggest to drop back to the version which supports older version of liburing and make a NOTE in the README about that situation.

I don't see this as feasible to maintain unfortunately.  Folly is one of many libraries from Meta that has a tight dependency chain amongst each other assuming alignment mostly within their infra and select distro CI pipelines.  In order to align to stable, we'd have to unwind to very old versions that are likely to be proven less useful w/ known isssues if there is even adoption/usage of this slackbuild. (The versioning is linear w/o formal branching so the weekly publishes are just a stamped out hash)

For these libraries, I will generally fork from `-current` to fulfill most recent dependencies which also walk back to more recent kernel versions as well.  With that said, when this slackbuild was initially submitted, it just so happened there was dependency alignment in order to build on `-stable` however I think we are well beyond that point now.

Slackbuilds are possible for this library and various others that depend on folly but imo, not on `-stable`

willysr:
Since that's the case, i will remove folly from SBo

PR #6208, opened 2024-04-29
development/rust16: Updated for version 1.77.2.

pghvlaans:
Sorry about the delay. Reverse dependencies (except for `yabridge`, which needs multilib) built on VM for 64- and 32-bit and in Docker for ARM.

The `sbopkglint` failure for 64-bit is expected; stripping shared objects causes a segfault when running the binaries.


willysr:
@sbo-bot: build development/rust16

willysr:
merged

PR #6246, opened 2024-05-02
system/openzfs: Remove unnecessary configure arguments.

mattegger:
Removed configure arguments that pointed to the "/lib/modules/${KERNEL}/source" symlink. These are not necessary to compile openzfs on -15.0, and cause the compile to fail on -current. As far as I can tell, the kernel-modules package from 15.0 creates this symlink, but the current version does not.


aclemons:
@sbo-bot: build system/openzfs

aclemons (approved):
LGTM

willysr:
merged

PR #6268, opened 2024-05-03
games/Mindustry: Repackage instead of Gradle build

Ythogtha:
At least it makes for an easier SlackBuild...

But it needs the icon manually added.
Maybe it could be extracted from the jar.


Ythogtha:
@sbo-bot: build games/Mindustry

willysr:
merged

PR #6296, opened 2024-05-04
python/pluggy: Removed (renamed to python2-pluggy)

isaackwy:
I already released a slackbuild named python2-pluggy (and all the dependencies such as tox and pytest link to python2-pluggy rather than pluggy).


willysr:
merged

PR #6316, opened 2024-05-07
multimedia/kodi uses local spdlog, fmt, and rapidjson & version bump multimedia/inputstream-adaptive to 21.4.6

bassmadrigal:
multimedia/inputstream-adaptive requires the inclusion of external rapidjson dependency to kodi, and kodi also switches to external spdlog and fmt since spdlog now requires fmt.


willysr:
merged

PR #6353, opened 2024-05-11
Disable building tests.

0xBOBF:
This commit disables building tests for mutter and gnome-shell, so that mutter can drop the dependency on xvfb-run. The gnome-shell slackbuild is included in this commit, since it must disable building tests as well, otherwise gnome-shell will fail-to-build when it tries to build its mutter tests.

I've tested building and runtime operation of the reverse dependencies of this change, which are gnome-shell, and the budgie-desktop packages. Everything is building and working as expected.

Edit: gdm and gnome-terminal are also building and running fine (along with all the other gnome packages I maintain). There's a bit of software to build for the whole dep chain so I wanted to add my results here if you want to go off that.


willysr:
@0xBOBF please split this into 2 commits, one for each script

0xBOBF:
> @0xBOBF please split this into 2 commits, one for each script

I'll close this PR and submit another with two commits. Having too much of a mess with my local branch being behind master now too, on-top of trying to split the commit.

PR #6375, opened 2024-05-12
Drop xvfb run

0xBOBF:
These commits disable building tests for mutter and gnome-shell, so that mutter can drop the dependency on xvfb-run. The gnome-shell slackbuild is included, since it must disable building tests as well, otherwise gnome-shell will fail-to-build when it tries to build its mutter tests.

I've tested building and runtime operation of the reverse dependencies of this change, which are gnome-shell, and the budgie-desktop packages. Everything is building and working as expected. gdm and gnome-terminal are also building and running fine (along with all the other gnome packages I maintain). There's a bit of software to build for the whole dep chain so I wanted to add my results here if you want to go off that.


willysr:
@sbo-bot: build desktop/gnome-shell

willysr:
@sbo-bot: build desktop/mutter

willysr:
merged

PR #6395, opened 2024-05-15
desktop/mousam: Update for 1.2.0

isaackwy:
I found that if mousam is installed through slpkg (rather than through sudo ./mousam.SlackBuild), the compiliation does not include the patches. This means that if mousam is installed through slpkg, then mousam does not run.


willysr:
@sbo-bot: build desktop/mousam

willysr:
merged

PR #6447, opened 2024-05-18
desktop/plasma5-themes-breath: Orphan maintenance of plasma5-themes-breath

isaackwy:
1. The Arch Linux AUR had already removed plasma5-themes-breath from their website; therefore, I do not have a consistent resource for making SlackBuild edits and package updates.
2. I don't use KDE - in fact, I uploaded plasma5-themes-breath to change my qt5 theme on non-KDE window/desktop environments. This means that I am not interested in testing plasma5-themes-breath installs (to see if the theme applies successfully).
3. I also fixed spacing within the "cmake" compilation section.


willysr:
merged

PR #6448, opened 2024-05-18
desktop/plasma5-themes-breath-migration: Orphan maintenance of plasma5-themes-breath-migration

isaackwy:
1. The Arch Linux AUR had already removed plasma5-themes-breath-migration from their website; therefore, I do not have a consistent resource for making SlackBuild edits and package updates. (This is similar to the situation with plasma5-themes-breath).
2. I don't use plasma5-themes-breath (therefore, by extension, I also don't use plasma5-themes-breath-migration).
3. I also fixed spacing within the "cmake" compilation section (similarly to the plasma5-themes-breath SlackBuild).


willysr:
merged

PR #6464, opened 2024-05-21
network/hiawatha: Fix build.

willysr:


willysr:
@sbo-bot: build network/hiawatha

PR #6479, opened 2024-05-22
network/zdns: Added (DNS Utility).

willysr:


willysr:
@sbo-bot: build network/zdns

PR #6528, opened 2024-05-26
audio/Ymuse: Edit commenting

isaackwy:
Regarding the command to install local language files:
I am unsure whether it "does not install anything so far", particularly on non-English locales.

Therefore, please remove the code comment.


willysr:
merged

PR #6631, opened 2024-06-01
ci: Show reverse deps as a tasklist.

aclemons:
This syncs the output for gitlab MRs to also be a tasklist like we
do for github PRs.


willysr:
merged

PR #6637, opened 2024-06-03
gis/gdal: Updated for version 3.9.0.

ArTourter:

Java binding requires JDK 11 or above. Tested with zulu-openjdk (11 & 17) and alien openjdk17.
poppler patch not required with version 3.9.0.


willysr:
@sbo-bot: build gis/gdal

willysr:
i think this requires numpy?

ArTourter:
> i think this requires numpy?

Huh! it is just advisable but now they have added a cmake option if one wants to compile without it. OK I'll add it to the list of reqs, check again the java reqs in case it was a red herring and send an update.

aclemons:
@sbo-bot: build i586 gis/gdal

ArTourter:
ok I have added numpy as a dependency. I don't see the point of making that optional or not including it.
I have also checked if the JAVA bindings compile with JDK8 and it does, so I have removed the note in the README file.

willysr:
@sbo-bot: build gis/gdal

ArTourter:
I am really sorry. I will get this right one day!

aclemons:
@sbo-bot: build gis/gdal

aclemons (approved):
LGTM

PR #6739, opened 2024-06-12
system/trash-cli: Add shell completions

isaackwy:
See the comments within https://github.com/SlackBuildsOrg/slackbuilds/pull/4306 for more details (that was when I orphaned python3-shtab).

It would be better if I could trigger a re-installation of trash-cli immediately after getting python3-shtab installed.

To get around the issue of shell completion support having to be added at compile-time (that is, running the SlackBuild with python3-shtab installed beforehand), rather than immediately after installing python3-shtab on a system with trash-cli, I worded the README the best I could.


willysr:
@sbo-bot: build system/trash-cli

willysr:
merged

PR #6740, opened 2024-06-12
development/electron: Update for 26.6.10

isaackwy:
While electron26 is end-of-life:

This version should still provide compatibility with the oldest of programs that still rely on Electron.


willysr:
@sbo-bot: build development/electron

willysr:
merged

PR #6804, opened 2024-06-17
desktop/mint-themes: Add python3-libsass build dependency

isaackwy:
I found out just now that python3-libsass is supposed to be an additional build dependency for mint-themes.
Otherwise: (ex. colors do not change upon switching to one of the Mint gtk2/gtk3 themes).


desktop/mint-themes/mint-themes.info
```diff
@@ -5,6 +5,6 @@ DOWNLOAD="https://github.com/linuxmint/mint-themes/archive/2.1.7/mint-themes-2.1
 MD5SUM="21f67918c100e7a860ac85b188637df0"
 DOWNLOAD_x86_64=""
 MD5SUM_x86_64=""
-REQUIRES=""
+REQUIRES="python3-libsass"
```
willysr, line 8: Actually it's pysassc
https://github.com/linuxmint/mint-themes/commit/b1dfbc54992cdcc8cb0c64e9bf5ac44d69ef9709

willysr: libsass seems to work as well

willysr:
merged

PR #6805, opened 2024-06-17
python/python3-pexpect: Update for 4.9.0 (+new maintainer)

isaackwy:
Benjamin Trigona-Harany is not maintaining SlackBuilds anymore.
I would like to take over maintenance of python3-pexpect.


willysr:
merged

PR #6814, opened 2024-06-18
system/mailfromd: Update for version 9.0

graygnuorg:

PR #6821, opened 2024-06-19
libraries/adns: Update for version 1.6.1 (Fixed DESTDIR)

graygnuorg:
This is an updated version of PR #I6815. It turned out that the original adns-1.6.1 mishandles DESTDIR when doing make install. This, in particular, caused build failures in #6814. This updated PR fixes it.


libraries/adns/adns.SlackBuild
```diff
@@ -88,15 +89,11 @@ CXXFLAGS="$SLKCFLAGS" \
   --mandir=/usr/man \
   --infodir=/usr/info \
   --docdir=/usr/doc/$PRGNAM-$VERSION \
-  --disable-static \
```
willysr, line 91: why are these lines are deleted?

libraries/adns/adns.SlackBuild
```diff
@@ -88,15 +89,11 @@ CXXFLAGS="$SLKCFLAGS" \
...
-  --enable-shared \
   --build=$ARCH-slackware-linux
 
 make
 make install DESTDIR=$PKG
 
-find $PKG -name \*\.a -type f -delete
```
willysr, line 98: also this one

willysr (commented):
These two lines should not be deleted as they ensure we build shared libraries

libraries/adns/adns.SlackBuild
graygnuorg: > why are these lines are deleted?

Because adns configure script does not understand them. With them you get:

libraries/adns/adns.SlackBuild
graygnuorg: No objections. Restored this one.

libraries/adns/adns.SlackBuild
willysr: ah ok

willysr:
@sbo-bot: build libraries/adns

willysr:
merged

PR #6858, opened 2024-06-23
system/realtek-bt: Align with template, respect $KERNEL.

aclemons:
Packages for kernel modules on SBo should respect an env var KERNEL. This allows installing new kernel packages, then building packages for modules with the new version before rebooting, installing those and then rebooting into the new kernel. Then if desired, the duplicate packages can be removed when everything is working.

This also means the kernel version should be part of the package version so you can install a package for both kernels.

I've also adjusted the doinst.sh/douninst.sh to operate on the module from the package.


aclemons:
@sbo-bot: build system/realtek-bt

aclemons:
@sbo-bot: build system/realtek-bt

willysr:
LGTM

willysr:
merged

PR #6860, opened 2024-06-23
network/pound: Fix permissions on /etc files.

aclemons:
Files from `$CWD` should not be installed with `cp`, but rather `install` so that have correct ownership and permissions.

### DIFF
--- a/network/pound/pound.SlackBuild
+++ b/network/pound/pound.SlackBuild
@@ -102,6 +102,9 @@ find $PKG -print0 | xargs -0 file | grep -e "executable" -e "shared object" | gr
 find $PKG/usr/man -type f -exec gzip -9 {} \;
 for i in $( find $PKG/usr/man -type l ) ; do ln -s $( readlink $i ).gz $i.gz ; rm $i ; done
 
+rm -f $PKG/usr/info/dir
+gzip -9 $PKG/usr/info/*.info*
+
 mkdir -p $PKG/usr/doc/$PRGNAM-$VERSION
 cp -a \
   AUTHORS COPYING ChangeLog NEWS README THANKS \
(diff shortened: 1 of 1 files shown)


aclemons:
@sbo-bot: build network/pound

network/pound/pound.SlackBuild
```diff
@@ -109,8 +109,8 @@ cp -a \
 cat $CWD/$PRGNAM.SlackBuild > $PKG/usr/doc/$PRGNAM-$VERSION/$PRGNAM.SlackBuild
 
 mkdir -p $PKG/etc $PKG/etc/rc.d
-cp $CWD/pound.cfg $PKG/etc/pound.cfg.new
-cp $CWD/rc.pound $PKG/etc/rc.d/rc.pound.new
+install -D -m 0644 -o root -g root $CWD/pound.cfg $PKG/etc/pound.cfg.new
+install -D -m 0755 -o root -g root $CWD/rc.pound $PKG/etc/rc.d/rc.pound.new
```
willysr, line 113: shouldn't rc.xxx be installed with 0644 by default? they need to enable (+x) it manually if it's needed

willysr (commented):
default permission?

network/pound/pound.SlackBuild
aclemons: I grepped all the slackbuilds and it seems some do 0644 some 0755. Is there an FAQ for which is "correct"?

willysr: i think 755 should only be binaries, but rc.xxx should be 644 by default

aclemons: I switched it to 644

aclemons:
@sbo-bot: build network/pound

willysr:
merged

PR #6913, opened 2024-06-26
Homepage updates

bassmadrigal:
Used the below command to review homepages for my packages. Any not listed as 200 were checked for validity. Those remaining are due to language or version specific redirects that aren't worth updating the main homepage listing (except python3-pyproject-api, since it seems their readthedocs site seems to be currently broken -- I may update this in the future if it's not fixed) or because I already have updates submitted for this week (they'll be queued for a future update). For those I noticed were missing HOMEPAGE in the slack-desc, the line was added with the correct homepage.

No changes occurred in the SlackBuilds or building process, so BUILD bumps are not needed and I don't believe rebuild attempts with sbo-bot should be done (but I can initiate them if desired).

`for i in $(grep $MYEMAIL */*/*.info | cut -d: -f1); do . $i; curl -s -w %{http_code} $HOMEPAGE -o /dev/null; echo " - $i - $HOMEPAGE"; done | grep -v ^200 | sort`


willysr:
merged

PR #6917, opened 2024-06-27
python/python3-mistune: Update README, +remove SLKCFLAGS and LIBDIRSUFFIX

isaackwy:
I am not updating python3-mistune to the latest current version (3.0.2). I don't know how such an update will affect jupyter-notebook, jupyterlab and spyder.

This is what I will do instead:
1. Update README. About a year ago, Yth has already removed python2-mistune from SlackBuilds.org. I would like to change the README accordingly.
2. Remove SLKCFLAGS and LIBDIRSUFFIX (these 2 variables are never used within the SlackBuild).


willysr:
merged

Issue #6944, opened 2024-06-29
15.0-20240629.1: Build check


aclemons:
#  Overview

|  | package | i586 | x86_64 | notes | resolution |
| --- | --- | --- | --- | --- | --- |
|  | network/frr | :x:  | :x:  | broken by protobuf upgrade  |  |
|  | system/pcp | :x: | :ok: | fails with multiple jobs on i586 | #6961 |
  
# Notes 

## Raw list x86_64

https://github.com/SlackBuildsOrg/slackbuilds/issues/6944#issuecomment-2198007990

## Raw list i586

https://github.com/SlackBuildsOrg/slackbuilds/issues/6944#issuecomment-2198142716

aclemons:
@sbo-bot: build x86_64 all

aclemons:
@sbo-bot: build libraries/ospray

aclemons:
Ah, this is the one that needs a manual build because it needs ispc which needs llvm-17.0

aclemons:
@sbo-bot: build i586 all

aclemons:
@sbo-bot: build i586 system/pcp

aclemons:
@sbo-bot: build amd64 system/pcp

aclemons:
I could build system/pcp on i586 by forcing a single job. Seems flaky with multiple jobs. It previously built fin on 15 June:

```sh
# ls -la
total 9184
drwxr-xr-x   2 root root    4096 Jun 29 13:32 ./
drwxr-xr-x 318 root root   12288 Jun 29 13:32 ../
-rw-r--r--   1 root root    1028 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.dep
-rw-r--r--   1 root root  178032 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.lst
-rw-r--r--   1 root root     403 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.meta
-rw-r--r--   1 root root     180 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.txt
-rw-r--r--   1 root root 9180444 Jun 15 09:21 pcp-6.0.4-i586-3_SBo.txz
-rw-r--r--   1 root root      59 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.txz.md5
-rw-r--r--   1 root root      91 Jun 15 09:22 pcp-6.0.4-i586-3_SBo.txz.sha256
[...]

PR #6961, opened 2024-06-30
system/pcp: Use a single build job on 32bit.

aclemons:
This causes random errors on 32bit. We have not seen the issue on
64bit so we will not force a single job there since this package
takes a while to build.

Related to #6944


aclemons:
@sbo-bot: build system/pcp

willysr:
maybe we need to change cp -a into cp or install in order to get the permission right?

willysr:
merged

PR #6968, opened 2024-07-01
python/python3-tenacity: Update for 8.4.2

isaackwy:
I don't think python3-trio is an additional runtime dependency since python3-tenacity 8.4.0.
Instead, python3-tenacity 8.4.0 now support python3-trio out of the box.


willysr:
@sbo-bot: build python/python3-tenacity

willysr:
merged

PR #6969, opened 2024-07-02
python/munch: Update for 4.0.0 (+new maintainer)

isaackwy:
I plan on eventually taking over geopandas (maybe next month).

1. Slackware 15.0 already installs python-PyYAML by default (therefore, I have edited the README removing the mention of munch making use of PyYAML and python3-PyYAML).
2. Munch does not require python2-pbr as a dependency (as Slackware 15.0 already has python-pbr built in Python 3).
3. Munch now requires a newer version of setuptools (that is, python3-setuptools-opt).


willysr:
@sbo-bot: build python/munch

willysr:
merged

PR #7019, opened 2024-07-06
desktop/mousam: Orphan maintenance of mousam [edit: Remove mousam as well]

isaackwy:
I would like to orphan mousam - I am no longer maintaining it.

I am not using mousam as a weather viewer.

EDIT: Thanks @willysr for the advice. I have changed the pull request accordingly (remove mousam from SlackBuilds.org).


willysr:
instead of orphaning, it would be better to remove it instead as we will remove any orphaned packages for next cycle anyway

willysr:
@isaackwy

willysr:
merged

PR #7043, opened 2024-07-07
desktop/breath-gtk-theme: Removed (broken and outdated GTK theme)

isaackwy:
I uploaded breath-gtk-theme during the Slackware 14.2 period (which contained older GTK versions).
Due to updates to GTK+2 and GTK+3, this theme is now broken.

As an example, I have displayed a Bleachbit window (a GTK3 interface):
Bleachbit_Window


willysr:
merged

PR #7093, opened 2024-07-09
libraries/rest: Remove unneeded deps.

willysr:


willysr:
@sbo-bot: build libraries/rest

PR #7112, opened 2024-07-11
academic/root: Updated for version 6.30/08, removed gsl_shared flag.

arabusov:


arabusov:
@sbo-bot: build academic/root

willysr:
@sbo-bot: build academic/root

willysr:
@arabusov this failed to build

arabusov:
@willysr what access rights are required for `/etc/ld.so.conf` and `/etc/ld.so.conf.d`? On my system I have 644 and 755, shall I set them explicitly in the Slackbuild?

willysr:
that's the correct permission

willysr:
i think you can mark it as x86_64 only

arabusov:
> that's the correct permission

Ok. But why the "sanity" test reports
```
--- forbidden directory exists: etc/ld.so.conf.d
--- forbidden file: etc/ld.so.conf
```

willysr:
you need to ask Urchlay for that. He is the one that build the sbolint and sbopkglint tool

arabusov:
> i think you can mark it as x86_64 only

Occasionally, I use TP X31, a very nice machine, and it would be nice to have ROOT there as well. I'll try to figure out why it fails with newer ROOT version. Btw, the ROOT team managed to compile it for 32 bit Windows: https://root.cern/releases/release-63008/

willysr:
i will merge this with x86_64 only for now.

PR #7135, opened 2024-07-13
development/bazel: Update for 5.4.1

isaackwy:
1. Install the files at /usr/share/bazel instead of /opt/bazel
2. bazel still should not be stripped (otherwise, it does not work as a build dependency for programs such as workerd and tensorflow)

Also, though the latest version of bazel is at 7.2.0, I do not plan on updating bazel any later than 5.4.1.
EDIT: I noticed that Andrew Clemons made this SlackBuild 64-bit. I rebased this PR with the change in mind.


willysr:
@sbo-bot: build development/bazel

willysr:
merged

PR #7136, opened 2024-07-13
audio/tauonmb: Let's clean this mess.

RSKYS:
> Too many dependencies already, a little cleanup.

I don't even know what to say, I really like this player, with a little bit cleanup, it's gonna be good again.
I understand they want to have most of compatiblity with everything but this is not vlc, and I'm not so happy with these many dependencies.
I also really want users to be in charge of changes, so if my use case is not gaming music deconding I don't want it, if I want it, I build it using LIBGME=yes

> Switched to Unidecode (seems python3-isounidecode is long dead).

I'm not sure if I should remove python3-isounidecode, I'll keep it around in case me or someone else will need it again later.


willysr:
@sbo-bot: build audio/tauonmb

willysr:
merged

PR #7188, opened 2024-07-19
audio/yabridge: Fix dbus problem

noisecode3:
This fixes a dbus connection problem some users seem to have and it also fix unsupported problem in info file


willysr:
@sbo-bot: build audio/yabridge

willysr:
merged

PR #7218, opened 2024-07-20
python/pyudev: Removed (rename to python3-pyudev)

isaackwy:
All other SlackBuilds that previously depended on pyudev now depend on python3-pyudev.


willysr:
merged

PR #7156, opened 2024-07-15
development/google-go-lang: Updated for version 1.22.5.

aclemons:
I'm not the maintainer here, but the admins have been keeping this
package updated since quite a few builds in the repo rely on it:

https://git.slackbuilds.org/slackbuilds/log/development/google-go-lang

These patch releases include quite a few security fixes:

https://go.dev/doc/devel/release#go1.22.minor

PR #7081, opened 2024-07-09
graphics/graphviz: updated for version 12.0.0

Ythogtha:

PR #6959, opened 2024-06-30
python/python3-gast: Version bump to 0.6.0

bassmadrigal:
@willysr, the version bump to 0.5.5 from #6910 seems to have been missed in the public update as well.

They released an even newer version, which this PR contains.

PR #7220, opened 2024-07-20
python/python3-gast: Downgrade to 0.5.5

bassmadrigal:
I didn't notice this update broke python/python3-pythran (and packages depending on that, like academic/python3-scipy). Currently, pythran only supports 0.6.x versions of gast in it's master branch. Downgrading to 0.5.5 until pythran on SBo supports newer versions of gast.

Thanks to Alexander Verbovetsky for the heads up!


bassmadrigal:
@sbo-bot: build python/python3-gast

bassmadrigal:
@sbo-bot: build python/python3-pythran

bassmadrigal:
@isaackwy, sorry I broke your python/python3-pythran SlackBuild and failed to notice it :wink:

willysr:
there's a reverse dep check on our CI, so whenever you pushed a new PR and see there's other scripts listed, please build them as well @bassmadrigal

aclemons:
Not so sidetrack the discussion, but this is one reason why I advocate for listing all direct dependencies in REQUIRES. There are at least two advantages:

- you can list first level dependencies and know what is required to build *this* package
- if the dependencies of your dependencies change, your build is not at risk of breaking because a dependency you need is no longer transitively available for your dependencies.

I know some of the others are advocates of "trimming" dependencies, i.e. not listing a dependency if a dependency itself needs that dependency, but I am not so much.

Here for example, the reverse dependency check does not list python3-pythran because it does list python3-gast as a dependency.

https://github.com/SlackBuildsOrg/slackbuilds/blob/84cbf1cf8caa7e0b9a925644b87da701377185eb/.github/workflows/ci.yml#L129-L135

It is actually though as you see here:

https://github.com/serge-sans-paille/pythran/blob/8a283ce2e9d407a28649097a1a951972badd7d04/requirements.txt#L3

It is relying on the fact that python3-beniget pulls in python3-pythran.

I actually think explicitly listing them is more in spirit of the guidelines anyway:

https://slackbuilds.org/guidelines/

> REQUIRES="(required build-time and run-time dependencies; optional dependencies at maintainer discretion)"

bassmadrigal:
@willysr, it [did not list pythran](https://github.com/SlackBuildsOrg/slackbuilds/pull/6959#issuecomment-2198484459) as a dependency. Even then, both pythran and beniget built fine with it, and you can even import the module using `python3 -c "import pythran"` with the broken gast version without error. For some reason, python3-scipy calls pythran during building in a way that shows pythran is broken. It's just something that slipped through the cracks and my normal checks missed it.

The only way sbo-bot would've found it would be to build reverse dependencies using "full" instead of "1-level", which seems like a lot of compilation time.

>whenever you pushed a new PR and see there's other scripts listed, please build them as well

Would this include something like python3-lxml (#6518), which has 26 packages relying on it?

When I've paid attention in the past, it didn't seem like you were building the reverse dependencies of my package submissions. If this is the intention, would it be beneficial to just have sbo-bot be automatically called when PRs are submitted and the reverse dependency check is complete with the submitted package and any reverse dependencies?

I also wasn't sure if the compilation time is provided by github for free or if there's a cost associated with it that's being paid by SBo or an individual. That uncertainty led me to keep the extra testing and compilations to my local system rather than be build heavy with sbo-bot.

aclemons:
> Would this include something like python3-lxml (#6518), which has 26 packages relying on it?

There is no perfect solution here and it mostly a subjective decision on our part. Some project take sem-version very seriously and we can trust them more to not break things, while others do not. But we are also not familiar with the release cycle and processes for all packages, so mistakes will happen.

In those cases, we get a report from our helpful users and it gets fixed.

> If this is the intention, would it be beneficial to just have sbo-bot be automatically called when PRs are submitted and the reverse dependency check is complete with the submitted package and any reverse dependencies?

I have written this before, but I will not do this. slackbuilds run as root and if this were the case, anyone could open a PR in this repo and have the build server run the script as root (it's run in a docker container started by non-privileged user, but still privileged in the container). The script could do anything, if someone wants to do something bad. We _always_ review the script before we build it.

I do want to add a new command though to build the reverse deps without having to list them, something like `@sbo-bot: build reverse category/package` or something like that.

> I also wasn't sure if the compilation time is provided by github for free or if there's a cost associated with it that's being paid by SBo or an individual. 

The sbolint checks and reverse dependency lookups that are posted as comments are done via a github action and those are free for public repos. The builds are run on a server I maintain and pay for with an AMD Ryzen 5 3600. It's a dedicated server, so it does not cost more to run more builds or anything, but it only runs one build at a time and the others are queued up since I have the runner setup to use all cores for each build.

Right now for example, it is rebuilding the whole repo after the public update, so your builds requested above are waiting in the queue (probably will take more than 10 hours to update everything this week with the mass replacements and fixes).

aclemons:
> Would this include something like python3-lxml (#6518), which has 26 packages relying on it?

I approved this one and did not rebuild since this was a patch release and also because I do python at work and we use python3-lxml and we'd already bumped this ourselves was quite sure it was fine.

willysr:
@bassmadrigal: some maintainers have been given access to the CI but not many were using it, while we have to deal with the rest of the PR and normal submissions.

I normally don't test Python scripts since most of it is minor changes, but I did test some other packages that I suspect have a breakage although some were false alarms.

We built this tool not just to help us, but maintainers as well since your scripts may have an impact on others, so please use them as much as possible :)

willysr:
if you want to see some PRs which i tested for all reverse deps:
- google-golang (https://github.com/SlackBuildsOrg/slackbuilds/pull/7156)
- graphviz (https://github.com/SlackBuildsOrg/slackbuilds/pull/7081)

isaackwy:
> @isaackwy, sorry I broke your python/python3-pythran SlackBuild and failed to notice it 😉

It's fine Jeremy.
I didn't rebuild python3-pythran so I didn't even notice it.

Note: I am not upgrading python3-pythran until a new release of Slackware comes out.
python3-scipy (which you are maintaining) cannot be updated in Slackware 15.0 anyway.

Newer versions of python3-scipy will need newer versions of python3-pythran, though...

willysr:
merged

bassmadrigal:
Ok, good. I'm glad that it doesn't cost extra for additional computational time. I can definitely be more proactive with testing. It also now makes perfect sense on why these builds aren't automatically triggered. Thanks @aclemons!

As for reverse dependencies, do we want every single one checked for every single submission or are we still able to be subjective and only do it when we think there might be a breaking change? Most of my updates are python based, and I'm usually able to catch issues locally (especially since they're usually runtime issues, not build issues).

Thanks for the clarification on the build system! It's definitely a great feature.

willysr:
It's up to you to decide which other packages should be tested after your changes. You have all the power to do so ;)

bassmadrigal:
Got it! Thanks!

PR #7235, opened 2024-07-22
audio/tauonmb: Maintainer's adjustment.

RSKYS:
Ctrl + R really enough for rename.
F2 key used to be theme changer in tauonmb old days, if you're old user and removed data folder for fresh start, you have probably realized. Now, current data of user won't be compromised; users with new data or those who wish to remove their data Should experience this key like old days :)


willysr:
merged

PR #7250, opened 2024-07-24
libraries/allegro: updated REQUIRES

Ythogtha:
It seems easier, here, to set physfs as REQUIRES instead of optional.
If physfs is present, allegro is automatically built against it, and one of the only three slackbuilds using allegro, namely opensurge, will require allegro to be built against physfs.

It feels like this is better to set physfs as a REQUIRES than to expect people to read the README of opensurge and then rebuild allegro from scratch after having installed physfs.


willysr:
merged

PR #7295, opened 2024-07-26
Yelpwebkit2gtk4.1

timsoftgit:
this adds behind the scenes support for webkit2gtk4.1 for yelp, and the packages that use it. xiphos, gnucash (indirectly via gnucash-docs). no changes required for gmdb2. Unless webkit2gtk4.1 is installed, there will be no changes to the package.
Note: switching to webkit2gtk4.1 is an all or nothing step, as a binary cannot mix soup2 and soup3 libs, hence the three updates.

The gnome project has obsoleted soup2 in favour of soup3, changing the api but keeping function names so that libs cannot be mixed in the same binary. As a result, many packages have not been updated by gnome devs, such as libgdata, gnome-photos, gnome-shell, gnome-terminal or grilo-plugins, gfbgraph, evolution-data-server etc. some of which have no maintainers, or have been abandoned in newer versions of other libs/apps. This has effectively broken the available gnome apps going forward.
As more applications abandon soup2/webkit2gtk4 in favour of soup3/webkit2gtk4.1 this provides a headache for a path forward for slackware.
These updates, while not affecting webkit2gtk users, at least provide a possible migration path for the future. (which is the intent).

### DIFF
--- a/academic/xiphos/README
+++ b/academic/xiphos/README
@@ -4,3 +4,6 @@ Sword project by the CrossWire Bible Society.
 
 computer reading of the text using festival has not been tested. if
 it works let me know, and i'll remove these two lines from the readme.
+
+if built with webkit2gtk4.1 then yelp also needs to be built with
+webkit2gtk4.1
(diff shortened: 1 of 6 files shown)


willysr:
I have merged this pr

timsoftgit:
thanks.

PR #7335, opened 2024-07-29
Typosfix

timsoftgit:
fix a couple of typos that stopped webkit2gtk4.1 detection working in xiphos and gnucash. (appologies for the original typos.)


willysr:
merged

PR #7341, opened 2024-07-29
Fix Flameshot installation slackbuild script

sanel:
When slackbuild script tries to install flameshot files, it will fail because 'cp' tries to copy folder 'docs' and for that it needs '-r' (recursive) option.


willysr:
i have it fixed already on my branch

PR #7350, opened 2024-07-31
audio/tauonmb: Update script (Build 3).

RSKYS:
New patch fixes actual UI bug, specifically the custom dark high contrast theme (Based on recent github commit).


willysr:
merged

PR #7369, opened 2024-08-01
python/python3-aiohttp: Updated for version 3.10.0.

mac-a-r0ni:
This now requires new approved upload dep "python3-aiohappyeyeballs" which has been added to REQUIRES.


willysr:
merged

PR #7371, opened 2024-08-01
system/virtualbox: fix error when starting

ardrabczyk:
As reported at https://lists.slackbuilds.org/pipermail/slackbuilds-users/2024-July/029553.html.

I found that modules are loaded and `VBoxManage list vms` doesn't
return non-zero even when the error is printed so maybe at the end of
the day it didn't prevent VirtualBox from working correctly but it
should be removed nonetheless.


willysr:
@pprkut can you take a look on this?

pprkut (approved):
Looks good to me. Thanks!

willysr:
merged

PR #7372, opened 2024-08-01
system/nix: Organize script.

RSKYS:
Hi Willy,
I finally figured this out, there should be no conflict now.
I thought better to submit it, it also has new syntax for cpp20 constructor, so it will become less annoying for users; they build the script without regular warns of it.

It's not necessary to go for build 2, it's simply just cleanups and improvements.


willysr:
@sbo-bot: build system/nix

willysr:
merged

PR #7406, opened 2024-08-03
python/python-evdev: Removed (renamed to python3-evdev)

isaackwy:
I now have renamed python-evdev to python3-evdev.
All the SlackBuilds that previously depended on python-evdev now depend on the renamed python3-evdev.


willysr:
merged

PR #7439, opened 2024-08-05
python/python3-aiohappyeyeballs: Cleanup SlackBuild script.

mac-a-r0ni:


willysr:
Removing credit is not something we should do as they wrote the original script

willysr:
merged

PR #7514, opened 2024-08-12
Fixed source download link

cduston44:
The source for this version was removed from the Insync repository (not great...), but there is a mirror at Slackware UK.


willysr:
merged

PR #7614, opened 2024-08-19
network/electrs: Added electrs - Electrum Server

thnkman:


willysr:
For new submission, please send it via submission form
https://slackbuilds.org/submit/

PR #7774, opened 2024-08-31
python/typing-extensions: Rename deps to python/python3-typing-extensions and remove typing-extensions

bassmadrigal:


bassmadrigal:
Should not require any rebuilds as the resulting package remained unchanged except for the package name.

willysr:
merged

PR #7780, opened 2024-08-31
development/spyder: Add further commenting

isaackwy:
I mentioned beforehand that python3-spyder-kernels 3.0.0 causes Spyder to crash.
I need to add that this applies to Spyder 5.4.0 (the version at SlackBuilds.org) specifically.

The crash might not happen with later versions of Spyder (but I have not tried said later versions yet).

Also, remove SLKCFLAGS and LIBDIRSUFFIX (these variables are not used in the SlackBuild).


willysr:
merged

PR #7788, opened 2024-09-01
network/dnscrypt-proxy - Quad9 Signing key has changed

thnkman:
Details: https://github.com/DNSCrypt/dnscrypt-proxy/discussions/2679


aclemons:
You probably want to bump `BUILD` for this - sbopkg users etc will notice the change then.

willysr:
merged

PR #7793, opened 2024-09-02
audio/yabridge: Fix info file

noisecode3:
Because it build both on 64 and 32 it should not be unsupported on "DOWNLOAD_x86_64", it should all be in "DOWNLOAD" and "MD5SUM" correct?


willysr:
well, not based on this build https://github.com/SlackBuildsOrg/slackbuilds/pull/7188
it failed to build on pure x86_64

PR #7794, opened 2024-09-02
development/azuredatastudio: Updated for version 1.49.1.

noisecode3:


development/azuredatastudio/azuredatastudio.SlackBuild
```diff
@@ -22,10 +22,10 @@
 #  OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
 #  ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-cd $(dirname $0) ; CWD=$(pwd)
+cd "$(dirname "$0")" || exit 1; CWD=$(pwd)
```
willysr, line 25: There's no need to use this exit 1 command as we already use set -e

willysr (changes requested):
please fix this

development/azuredatastudio/azuredatastudio.SlackBuild
noisecode3: "set -e" is used later on line 55, so you must mean that sbopkg are using "-e" to run the scripts? Are you sure you want me to change it?

willysr: that command has low probability of failing so there's actually no need to add exit 1
the one which has higher chance of failing is in the section where it builds and packaging the files/directory into the final output. That's why set -e is positioned after this part

noisecode3: I guess the only bad thing that can happen is that it will try to look for the tar source in current directory that comes from pwd. If someone did not cd into some safe directory where you keep those, it could be trying with some other tar. It's probably fine and I'm paranoid.

willysr:
@sbo-bot: build x86_64 development/azuredatastudio

willysr:
merged

PR #7859, opened 2024-09-06
libraries/libnfs: Version bump to 5.0.3 + new maintainer

bassmadrigal:


bassmadrigal:
@sbo-bot: build libraries/libnfs

bassmadrigal:
@sbo-bot: build multimedia/vlc

bassmadrigal:
Can anyone confirm whether the 32bit build fails with the original version of libnfs? My 32bit VM always locks up on me (haven't been able to figure out why).

willysr:
@cwilling can you test this PR on 32 bit?

cwilling:
Both original 5.0.1 and updated 5.0.3 build OK in up to date 32bit VM.
Here, the new 5.0.3 version causes the same vlc build error on 32bit, as above, but using the original 5.0.1 libnfs enables libnfs_plugin.la to compile without error and vlc builds to completion.

bassmadrigal:
@cwilling, thanks for looking into it. Guess I'll revert the update for now and just change the maintainer info until I can have more time to look into it.

cwilling:
Looking at vlc git repo, I see the most recent reference to libnfs is an update to use 5.0.2. I've just tried that version to build vlc and all looks OK now, so looks like a bump to 5.0.2 would be in order.

willysr:
that was almost 2 years ago
https://github.com/videolan/vlc/commit/e9526e8fc511158a40b85943817b0aad2b13cdf3

bassmadrigal:
@sbo-bot: build libraries/libnfs

bassmadrigal:
@sbo-bot: build multimedia/vlc

bassmadrigal:
@willysr @aclemons 

It's been resubmitted with 5.0.2. Hopefully VLC builds on 32bit fine this time.

willysr:
@cwilling @bassmadrigal should we worried about this output?
`usr/lib64/vlc/vlc-cache-gen: error while loading shared libraries: libvlc.so.5: cannot open shared object file: No such file or directory`

willysr:
merged

cwilling:
It's been a while since I was vlc maintainer but, from memory, vlc-cache-gen is being run from doinst.sh. However symlinks to .so files have not yet been generated (see /var/lib/pkg/scripts/vlc-3.0.20-...).

Despite that, it all somehow works on a running system or a full VM.  Maybe it's coming up as an error here due to the container/chroot environment being used - I see that the error occurs harmlessly at the end of `buildpkg` but runs OK at installation in `sbopkglint`

Maybe the doinst.sh.in could have "|| true" added to silence the error message i.e.
```
usr/lib%LIBDIRSUFFIX%/vlc/vlc-cache-gen usr/lib%LIBDIRSUFFIX%/vlc || true
```

PR #7905, opened 2024-09-07
python/pandocfilters: Removed (renamed to python3-pandocfilters)

isaackwy:
I now have replaced pandocfilters with python3-pandocfilters.

P.S. I should have named this branch pandocfilters rather than python3-pandocfilters. This also applies to the python3-terminado and python3-testpath branches (should have been named terminado and testpath respectively). However, that's just a minor detail.


willysr:
merged

PR #7906, opened 2024-09-07
python/terminado: Removed (renamed to python3-terminado)

isaackwy:
I now have renamed terminado to python3-terminado.


willysr:
merged

PR #7907, opened 2024-09-07
python/testpath: Removed (renamed to python3-testpath)

isaackwy:
I now have renamed testpath to python3-testpath.


willysr:
merged

PR #7908, opened 2024-09-07
Rename pytz dependencies

isaackwy:
I am renaming all the pytz dependency mentions.
This time, I am not bumping build version numbers (similar to Jeremy Hansen's approach to renaming typing-extensions dependency mentions).

In addition:
1. For the pandas SlackBuild, I am also renaming "python setup.py" to "python2 setup.py".
2. For the python3-matplotlib SlackBuild (which itself does not depend on pytz as a required dependency), I am renaming the mention of pytz (as an optional test dependency) in the README.


willysr:
merged

PR #7910, opened 2024-09-07
Update MAINTAINER and EMAIL fields in .info files of orphaned scripts

arcctgx:
These scripts don't have new maintainers after this week's update, so they are marked as orphaned, as agreed in the SlackBuilds-users mailing list.


willysr:
merged

PR #8022, opened 2024-09-14
python/pytz: Removed (has split into python2-pytz and python3-pytz)

isaackwy:
I now have delinked pytz as a dependency of other SlackBuilds (i.e. other SlackBuilds now depend only on python2-pytz or python3-pytz).


willysr:
merged

PR #8023, opened 2024-09-14
desktop/mint-themes: Revert download source to Linux Mint's repository

isaackwy:
I would like this SlackBuild to download the source from Linux Mint's repository rather than from the distribution's github mirror (i.e. the SlackBuild's previous behaviour before I took over maintenance).

I am not bumping the build version number (I am only changing the download source website).


willysr:
@sbo-bot: build x86_64 desktop/mint-themes

willysr:
merged

PR #8114, opened 2024-09-24
system/libratbag: Update for 0.18

isaackwy:
I don't know how to edit the SlackBuild such that /usr/lib/liblur.so.3.0.3 is instead installed in the /usr/lib64 folder.
For context, I use a 64-bit operating system.

Here is the relevant sbopkglint output:
```
Running test: 20-arch...
--- shared lib(s) in wrong dir for ARCH:
-rwxr-xr-x 1 root root 61832 Sep 24 13:45 usr/lib/liblur.so.3.0.3
___ note: 64-bit x86_64 package contains /usr/lib; this may or may not be a problem.
FAILED
```


willysr:
@sbo-bot: build system/libratbag

willysr:
merged

PR #8129, opened 2024-09-25
rust-opt reverse dependencies

pghvlaans:
This is everything besides `lean-elan` and `wezterm`, so far as I know.


willysr:
merged

PR #8368, opened 2024-10-13
audio/quodlibet: Update README file.

nick8325:
(there is no longer an optional dependency on webkit2gtk4 since the lyrics plugin has been removed)


nick8325:
Sorry, forgot to run sbopkglint! Looks like the build script needs a couple of fixes. I'll resubmit when it's done.

PR #8457, opened 2024-10-21
python/python3-diff-match-patch: Update for 20241021

isaackwy:


willysr:
@sbo-bot: build python/python3-diff-match-patch

willysr:
merged

Issue #8379, opened 2024-10-14
lua migration plan


willysr:
Based on https://gitlab.com/SlackBuilds.org/slackbuilds/-/issues/5, we will try to setup a migration plan for Lua:
- [ ] Submit new script "versioned" lua51 with versioned installed libs, includes and pkgconfig files
- [ ] Update Lua to 5.4 with the similar script as in the current and move as many scripts as possible to use Lua (5.4) or revert to a previous working version of Lua (51, 52, 53)  (WIP in https://git.slackbuilds.org/slackbuilds/log/?h=lua54-fixes)

ideas? @Ponce @aclemons

Ponce:
**IMHO** that's not the best way to go: the path to follow, like I wrote on IRC should be:
- lua on SBo should be updated to 5.4 using as a base the (unversioned) SlackBuild that is in current in a dedicated branch **(we shouldn't merge this branch until the next two steps are completed)**: we should migrate to that sooner or later so we should keep as much as possible compatible with lua in current, it makes no sense to do the same work twice;
- migrate the lua script presently on SBo to a versioned one (lua51, with accordingly installed libs and pkgconfig files);
- test what works with the newer lua (5.4.x) and edit the REQUIRES variable of whatever doesn't work with a the last working lua version (lua53, lua52, lua51): we should refer to archlinux's database to check the version used there and so avoiding doing this work twice.

FWIW it doesn't hurt to keep lua52 and lua53 in our repository, scripts that might need them could come up, sooner or later.

willysr:
ok, it's now updated to reflect your ideas
once it's settled, we can start working on it

willysr:
Progress: https://slackware.uk/~urchlay/src/lua54.notes.txt

pghvlaans:
RE Pioneer: Making it use Lua 5.4 is almost certainly impossible. The bundled Lua 5.2 works fine anyway, so I'll just remove the reference and detection-related code and submit a PR for this coming weekend.

pghvlaans:
@willysr Sorry, I misunderstood. `pioneer` builds and functions normally with Lua 5.4 on the system (or no Lua at all). I thought you were talking about getting it to *use* 5.4, which isn't possible.

willysr:
this should be closed

PR #8480, opened 2024-10-24
games/pioneer: Remove lua52 as an optional dependency.

pghvlaans:
RE #8379

Although the upstream default build uses bundled `lua-5.2`, I figured there wasn't any harm leaving `lua52` in place as an optional dependency so long as it was in the repo. Now that it won't be for much longer, it's time to drop it.


pghvlaans:
Good deal. Closing.

PR #8613, opened 2024-11-03
development/jupyterlab: Update for 4.0.13

isaackwy:
1. Since jupyterlab_server can be updated beyond version 2.16.3 on Slackware 15.0, I would like jupyterlab to be updated to 4.0.13 as well.

2. Remove the comment "JupyterLab will eventually replace the classic Jupyter Notebook" from the README and .info files. I don't think JupyterLab will replace Jupyter Notebook - that notice was written when JupyterLab was still at the version 3.x series.


willysr:
@sbo-bot: build development/jupyterlab

willysr:
merged

PR #8644, opened 2024-11-06
development/QtPy: Updated for version 2.4.2.

nomnombtc:
Hello,

for some reason the download link for this new version changed and it only
 works in lower case. So I've added SRCNAM=qtpy to the SlackBuild.


willysr:
@sbo-bot: build development/QtPy

willysr:
merged

PR #8733, opened 2024-11-13
desktop/i3: Update version to 4.24

emman:


willysr:
@sbo-bot: build desktop/i3

willysr:
@emman this failed to build

willysr:
@aclemons is this can be solved by adding hint?

aclemons:
Unfortunately not right now. Some things we just can't automate and have to manually check. I don't want to run the builds outside of a chroot (without the overlayfs) since I don't want trust running random scripts through the build server any more than we already do.

The only other idea I have would be if we could have a flag for running the tests which defaults to "on", but we set them not to run for the build checks.

willysr:
@emman can you add those flags?

willysr:
it's quite strange as previous changes works just fine
7ea3246d60b76d4e32321daf9e0e5c3e

willysr:
another way is to disable make test on perl section

aclemons:
Could only check on this. I tried this in an non-chroot env and it still fails.

I also failed to build it with sbopkg. Looking closely at the output, it is not because of the chroot, but rather because the build server is headless.

```
[...]
Calling i3 --get-socketpath failed. Is DISPLAY set and is i3 in your PATH? at /tmp/SBo/i3-4.24/AnyEvent-I3/blib/lib/AnyEvent/I3.pm line 142.
[...]
```

Maybe instead of the flag we talked about above, maybe we could just check if DISPLAY is empty and skip the tests in that case?

emman:
Hi, willy and aclemons, thanks for the feedback and help. 
I tested it with slackrepo and CHROOT enabled, and it worked as expected. The addition of the DISPLAY variable control was sufficient to resolve the issue. I also performed tests in a headless terminal environment, and everything functioned correctly.

aclemons:
@sbo-bot: build desktop/i3

aclemons (approved):
LGTM

aclemons:
staged to my branch

PR #8793, opened 2024-11-19
python/python-parse: Upgrade to 1.20.2.

henesy:
Maybe name change to python3-parse? 

True latest version requires a setuptools version higher than 15.0 offers, 1.19.0 seems to be the last setup.py version before the .toml switch.


willysr:
@sbo-bot: build python/python-parse

willysr:
merged

PR #8817, opened 2024-11-20
development/spyder: Fix requirements typo

isaackwy:
spyder 5.4.0 by default sets the qtconsole version requirement to be >=5.4.0, <5.5.0.

In line 87 of the spyder SlackBuild, I currently have that requirement changed to qtconsole >=5.5.0.
This is a typo.
Instead, that requirement is supposed to be: qtconsole >=5.4.0.


willysr:
merged

PR #8883, opened 2024-11-23
system/OpenSnitch: Edit commenting

isaackwy:
The link to pentoo's ebuild no longer works (it's for OpenSnitch 1.6.4).
I would like that link changed to refer to pentoo's ebuild for OpenSnitch 1.6.6-r1.


willysr:
merged

PR #8885, opened 2024-11-23
development/jupyter-nbconvert: Edit SlackBuild

isaackwy:
jupyter-nbconvert's default behaviour is to download the .css files if they're not already available in the source tarball.

The sed command to prevent downloading of .css files was only necessary for earlier versions of jupyter-nbconvert (as back then, the .css files were not provided in the tarball).

However, in later versions of jupyter-nbconvert (ex. jupyter-nbconvert 7.16.4), the .css files are already in the source tarball.

Therefore, I can remove the code section.
In fact, gentoo's ebuilds (for jupyter-nbconvert) have not listed this sed command since version 7.14.1.


willysr:
merged

PR #8915, opened 2024-11-27
desktop/mint-y-icons: Update for 1.7.8

isaackwy:


willysr:
@isaackwy please use github as packages.linuxmint.com does not retain outdated binary packages
it's safer to use github https://github.com/linuxmint/mint-y-icons/
same goes with other Cinnamon-based sources

PR #8966, opened 2024-12-01
development/yabasic: Updated for version 2.90.5

antonioleal:


development/yabasic/tar.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
willysr, line 1: i don't think this should be included in SBo?

willysr (commented):
it's better to have this removed from this PR

development/yabasic/slack-desc
```diff
@@ -16,4 +16,4 @@ yabasic: Yabasic runs under Unix and Windows and has a comprehensive
 yabasic: documentation; it is small, simple, open-source and free.
 yabasic:
 yabasic: Yabasic is very well documented.
-yabasic:
+yabasic: Homepage at http://2484.de/yabasic/
```
willysr, line 19: please don't put HOMEPAGE in README

willysr (changes requested):
There shouldn't be any URL in the README as well

development/yabasic/slack-desc
antonioleal: I actually used the URL in the slack-desc as part of my workflow, but ok, I'll arrange something different...

antonioleal:
Ok hope its alright now. I'm upgrading my workflow with new scripts, so some bugs might slip-in...

willysr:
@sbo-bot: build development/yabasic

willysr:
merged

PR #9044, opened 2024-12-08
system/doublecmd-qt5: New maintainer

antonioleal:
I'm taking over this since the maintainer does not respond to emails since a few month


willysr:
merged

PR #9105, opened 2024-12-14
development/electron: Update for 33.3.0

isaackwy:
I don't see any reason for holding back the electron version. I have not found any specific program that requires an older version of electron (except for pencil3, but that program is already difficult to get running anyway).

Also:
1. Fix typo in README (version in README should correspond with installed Electron version)
2. The SlackBuild should also create a /usr/share/electron symlink (I should not have removed it during a previous update).


willysr:
@sbo-bot: build x86_64 development/electron

willysr:
merged

PR #9106, opened 2024-12-14
development/spyder: Edit README

isaackwy:
More specifically specify that newer versions of spyder require python3-lsp-server >= 1.7.0.

Other users/maintainers do not know that python3-lsp-server v1.6.0 is the reason why spyder is being held back at v5.4.0.


willysr:
merged

PR #9107, opened 2024-12-14
python/python3-lsp-server: Edit README

isaackwy:
More specifically specify that newer versions of python3-lsp-server require python3-rope > 1.2.0.

Currently, python3-rope is being held back at version < 1.2.0.
Edit README to better reflect this context.


willysr:
merged

PR #9596, opened 2025-02-05
add numjobs to libqtpas

lecramyajiv:
Based on the openjdk21 slackbuild I have added the $numjobs to make.

### DIFF
--- a/libraries/libqtpas/libqtpas.SlackBuild
+++ b/libraries/libqtpas/libqtpas.SlackBuild
@@ -112,11 +112,22 @@ find -L . \
  \( -perm 666 -o -perm 664 -o -perm 640 -o -perm 600 -o -perm 444 \
   -o -perm 440 -o -perm 400 \) -exec chmod 644 {} \;
 
+nprocr=$(nproc)
+
+if [ -n $nprocr  ]; then
+    njobs="-j$nprocr"
+else
+    nv="$(expr $(nproc) + 1)"
+    njobs="-j$nv"
+fi
+
+echo "numjobs:$njobs"
+
 qmake-qt5 "QT += x11extras" \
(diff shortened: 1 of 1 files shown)


cwilling:
> Based on the openjdk21 slackbuild I have added the $numjobs to make.

I believe SBo generally prefers to use MAKEFLAGS rather than NUMJOBS. The OpenJDK21 maintainer claims that the use of MAKEFLAGS doesn't work in that SlackBuild, hence the use of NUMJOBS in that case.

Does MAKEFLAGS not work in this case and therefore require NUMJOBS?

willysr:
@lecramyajiv i used this on my own machine (/etc/profile)
`export MAKEFLAGS="-j 8"`

it works well, so I guess you don't really need to make any changes in libqtpas unless it doesn't honor as @cwilling said

lecramyajiv:
@willysr. I don't think everyone of the users will have explicitly added the MAKEFLAGS variable to their /etc/profile. 
so in the  libqtpas slackbuild i added these and checked it worked. export MAKEFLAGS="-j$(nproc)" and running make $MAKEFLAGS.  when running make alone without any parallel jobs it always assume "-j1" even if we have more than one cores. In my system without any makeflags or numjobs it takes about  roughly 20-25 minutes with numjobs  it reduces by half. Can i add export MAKEFLAGS="-j$(nproc)"  and make $MAKEFLAGS to the slackbuild

willysr:
you don't need to make changes in the SlackBuild with MAKEFLAGS, it will use whatever value you have in your local system, so if other people don't set it on their system, it will default to 1, where you can use whatever value you like in your system. This can be done without making any changes to the script

lecramyajiv:
ok ... I'll just leave the script as it is.

PR #9638, opened 2025-02-07
desktop/ClamAV-GUI: Updated for version 20250206_11ba095

antonioleal:


willysr:
@sbo-bot: build desktop/ClamAV-GUI

willysr:
@antonioleal this still failed

antonioleal:
I don't understand why it fails to extract the tar file.. on my qemu box the slackbuild works fine.

willysr:
you will have to submit via submission form for renaming, but I suggest that you change the source instead
Just like what I did with letsencrypt when they rename it to certbot, I kept using letsencrypt

PR #9667, opened 2025-02-10
development/zope.interface: Fix build when python-opt are present

Ythogtha:
If various python-opt packages are installed, zope.interface can only build if they are used.
I don't know precisely where lies the conflict, but with this modification it builds alone on a stock Slackware, or alongside lots of python builds.

This is required for things like python3-tubes which requires -opt packages for other dependencies, and will prevent zope.interface from building if other requirements are present...


Ythogtha:
Closing that one, batch updating, same fix.

PR #9656, opened 2025-02-09
python/python3-tubes: Fix build.

willysr:

PR #9668, opened 2025-02-10
Python3 tubes and dependencies

Ythogtha:
With those, python3-tubes seems to build properly, at last.
As no versions have been updated, only the build workflow is impacted, I'm testing only the top one : python3-tubes.
If that one builds with all its requirements, it will mean that all are building fine...

### DIFF
--- a/development/zope.interface/zope.interface.SlackBuild
+++ b/development/zope.interface/zope.interface.SlackBuild
@@ -27,7 +27,7 @@ cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=zope.interface
 VERSION=${VERSION:-7.2}
-BUILD=${BUILD:-2}
+BUILD=${BUILD:-3}
 TAG=${TAG:-_SBo}
 PKGTYPE=${PKGTYPE:-tgz}
 
@@ -65,6 +65,9 @@ find -L . \
   -o -perm 511 \) -exec chmod 755 {} \; -o \
  \( -perm 666 -o -perm 664 -o -perm 640 -o -perm 600 -o -perm 444 \
   -o -perm 440 -o -perm 400 \) -exec chmod 644 {} \;
+
+PYVER=$(python3 -c 'import sys; print("%d.%d" % sys.version_info[:2])')
+export PYTHONPATH=/opt/python$PYVER/site-packages
 
 python3 setup.py install --root=$PKG
 
--- a/python/python3-calver/python3-calver.SlackBuild
+++ b/python/python3-calver/python3-calver.SlackBuild
@@ -27,7 +27,7 @@ cd $(dirname $0) ; CWD=$(pwd)
 PRGNAM=python3-calver
 SRCNAM=${PRGNAM#python3-*}
 VERSION=${VERSION:-2022.6.26}
-BUILD=${BUILD:-1}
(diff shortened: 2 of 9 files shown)


Ythogtha:
@sbo-bot: build python/python3-tubes

willysr:
perhaps

willysr:
@Ythogtha please have a look on the issue list
I think most of the scripts here built fine without these changes

willysr:
see here https://github.com/SlackBuildsOrg/slackbuilds/pull/9656

Ythogtha:
Ok, my problem seems to be the order of builds.
Because I build dependencies of python3-tubes and installs them, but I don't isolate all dependencies builds.
So if I've installed, for example python3-automat or python3-incremental, then all the packages here are failing to build.
But if I starts with zope.interface, and python3-twisted, there won't be the build requirements of the other ones, and they'll build fine.

That's a bit unstable for me, it means some will only build on a clean python environment, and build order will matter for complicated packages, like buildbot, mailman, Kallithea, and probably a lot of others.

So the questions becomes : is it good to handle both environments, with or without -opt build packages, or should we do only what is required for building on a clean environment ?

willysr:
i simply used the queue from sbopkg, the same what our CI did

```
python3-characteristic
python3-pyasn1
python3-pyasn1-modules
python3-semantic-version
python3-flit_core
python3-installer
python3-pyproject-hooks
python3-build
python3-typing-extensions
python3-wheel
python3-packaging-opt
python3-setuptools-opt
python3-setuptools-rust-opt
rust-opt
python3-maturin
cryptography
python3-calver
python3-trove-classifiers
python3-pluggy
python3-pathspec
python3-editables
python3-hatchling
python3-hatch_vcs
python3-hatch_fancy_pypi_readme
python3-attrs
python3-service-identity
zope.interface
python3-tomli-opt
python3-incremental
python3-hyperlink
python3-versioneer
python3-constantly
python3-automat
python3-twisted
python3-tubes

```

Ythogtha:
I've narrowed the problem down to python3-incremental.
If it is built early, then zope.interface, python3-pyasn1, python3-pyasn1-modules, python3-characteristic, python3-calver, python3-pluggy, python3-trove-classifiers and python3-semantic-version will all fail to build, complaining that :
`pkg_resources.VersionConflict: (setuptools 57.5.0 (/usr/lib64/python3.9/site-packages), Requirement.parse('setuptools>=61.0'))`

So it's not because of the -opt packages, but strictly because of incremental...
I'm guessing sbopkg is lucky not to put it earlier in the build chain.

Ythogtha:
I'm pushing a revert on incremental, it - again it seems - breaks things.
Closing here.

Ythogtha:
But now Twisted is broken...

So I'm faced with :
- reverting incremental and twisted -> I don't want that especially...
- pushing this change here, which allows for all the packages here to build with the newer incremental -> looks like the safer option, but very intrusive...
- doing nothing else than hinting that incremental may prevent things to build, therefore requiring it to be removed, or user to export unilaterally PYTHONPATH=/opt/python$PYVER/site-packages for the builds (that works also).

What's your take on this ?

Ythogtha:
Closing here, I'm going for the information in incremental README.

PR #9669, opened 2025-02-10
python/python3-incremental: reverted to version 22.10.0 - breaks builds

Ythogtha:
Newer version will break a lot of builds, which would start complaining about a too old setuptools.
Reverting to avoid breakages, or lots of updates.

/!\ edit /!\ But breaks Twisted...


Ythogtha:
And *damn* that breaks the twisted update, which requires the newer incremental !

PR #9958, opened 2025-03-03
add xosview

r1w1s1:
add xosview 1.24


willysr:
@r1w1s1 please use normal submission for new script

PR #10085, opened 2025-03-13
multimedia/popcorntime: Updated for version 0.5.1

fsLeg:
Please use this submission instead of the one submitted through the site   (#10052). I didn't know the script polluted the filesystem so much until I saw the build log here. This should be amended now, except for Python scripts compiling and leaving behind .pyc files on 32 bit systems (that's just how Python works), but I'm not sure if this can or should be disabled.


willysr:
@sbo-bot: build multimedia/popcorntime

PR #10095, opened 2025-03-14
development/android-tools: Fix build with new profobuf3

philipvdh:
This fixes building with protobuf3 version 30. It also seems to work fine with 29.3 so it should not cause any issues in case this might get merged before the protobuf upgrade.


aclemons:
@sbo-bot: build development/android-tools

aclemons (approved):
LGTM

Issue #10021, opened 2025-03-07
protobuf 30.0 break


willysr:
This issue is used to track the progress with protobuf 30.0 (https://github.com/protocolbuffers/protobuf/releases) fixes for scripts that depends on it:
#  Overview
| package |  notes | resolution |
| --- | --- | --- |
| development/android-tools | --- | Fixed in #10095  | 
| gis/qgis | See https://github.com/protobuf-c/protobuf-c/pull/762 | Fixed by patching with https://github.com/protobuf-c/protobuf-c/pull/762 |
| libraries/grpc | --- | Updated for version 1.71.0 |
| libraries/protobuf-c | --- | Fixed by patching with https://github.com/protobuf-c/protobuf-c/pull/762 |
| network/feather | --- | Disable Trezor Hardware by default |
| python/python3-grpcio | -- | Updated for version 1.71.0  |
| system/usbguard | --- | Fixed with 3c46271be2b11281703baa1e13c07fd13074b885 |

willysr:
@isaackwy can you confirm that python3-grpcio is removed upstream? see notes

isaackwy:
No, python3-grpcio is not removed upstream.

My python3-grpcio Slackbuild is only the Python 3 build of gRPC: https://github.com/grpc/grpc
Recall that I split gRPC into the grpc and python3-grpcio SlackBuilds (because I wanted to make OpenSnitch maintenance easier. For context, OpenSnitch depends on python3-grpcio.)

I'm not using Slackware anymore so I can't test protobuf3 30.0 with python3-grpcio.

I have a suggestion: Update python3-grpcio to 1.71.0 (just like with grpc) and see what happens.

willysr:
@isaackwy 
python3-grpcio failed to build
https://github.com/SlackBuildsOrg/slackbuilds/pull/10020#issuecomment-2707904626

grpc built just fine though

isaackwy:
I think what https://github.com/protocolbuffers/protobuf/commit/5ba74b11e8d2bd5e9b22e972beb572668bf6191c is saying is that protobuf no longer provides Python RPC services by itself. Meaning, python3-grpcio (or rather, gRPC) must implement its own code generator plugins.

I don't know how this affects my python3-grpcio SlackBuild.
I looked at the build failure output, and it seems that grpc_tools (i.e. the python3-grcpio_tools that is bundled within the SlackBuild) is unable to find the generator.h file.

willysr:
[...]

PR #10020, opened 2025-03-07
development/protobuf3: Updated for version 30.0.

willysr:

PR #10109, opened 2025-03-15
python/python3-grpcio: Re-enable python3-grpcio-tools module [revised PR 1]

isaackwy:
See https://github.com/SlackBuildsOrg/slackbuilds/issues/10021 and https://github.com/SlackBuildsOrg/slackbuilds/pull/10020 for more details.

1. Can the python3-grpcio SlackBuild simply link to a prior protobuf v30.0 source tarball within the $TMP directory (that is, /tmp/SBo)? Or does the python3-grpcio SlackBuild itself have to download and extract another protobuf v30.0 tarball? (i.e. python3-grpcio's .info would additionally contain a link to the protobuf v30.0 source tarball.)

2. Regarding the protobuf version (the $PROTOVER variable):
Since @willysr prefers to keep updating protobuf, I am more inclined to remove the $PROTOVER variable entirely. Instead, the new build flag will be something like -I$TMP/protobuf-*/src

@willysr, please advise.

Update:
I'm trying this PR again.
This time, the python3-grpcio SlackBuild will download the protobuf v30.1 tarball and then extract it inside the grpc source folder (i.e. inside /tmp/SBo/grpc-1.71.0).

### DIFF
--- a/python/python3-grpcio/python3-grpcio.SlackBuild
+++ b/python/python3-grpcio/python3-grpcio.SlackBuild
@@ -16,7 +16,7 @@
 #  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO
 #  EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 #  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
-#  PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; 
+#  PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
 #  OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 #  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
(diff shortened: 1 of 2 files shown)


willysr:
@sbo-bot: build python/python3-grpcio

willysr:
@sbo-bot: build x86_64 system/OpenSnitch

willysr:
hm.. i updated protobuf to 30.1 on my branch, so this PR needs to be updated a bit
second, i don't know if it can look on other script's source directory (protobuf3) inside CI
may have to ask @aclemons for this

willysr:
@isaackwy i think python3-grpcio should be merged with grpc
that should work under CI i guess

willysr:
@sbo-bot: build x86_64 python/python3-grpcio

willysr:
i think you are not allowed to use other source's directory in CI

isaackwy:
> i think you are not allowed to use other source's directory in CI

What do you mean by this?

This is what we tried so far:
- set an additional build variable -I, linking it to /tmp/SBo/protobuf-30.1/src
- same thing as above, but instead extract protobuf-30.1 inside /tmp/SBo/grpc-1.71.0 (and then setting the build variable -I$TMP/grpc-1.71.0/protobuf-30.1/src)

willysr:
it requires a private headers found in protobuf package (see the patch i applied in protobuf to enable private headers).

in the PR, i don't see anything about second point, just first one

isaackwy:
> it requires a private headers found in protobuf package (see the patch i applied in protobuf to enable private headers).
> 
> in the PR, i don't see anything about second point, just first one

I'll add in the "re-enable private headers" patch tomorrow or so.
The build variable will also link to -I$TMP/grpc-1.71.0/protobuf-30.1/src, but the protobuf-30.1 source tarball (that I extracted to grpc-1.71.0) will be patched.

willysr:
it's better to put the PR in a complete state so i may not misunderstood your ideas

willysr:
merged in #10117

PR #10115, opened 2025-03-15
git/ci: Replace tj-actions/changed-files.

aclemons:
This repo has been deleted after it was compromised. We do not have
any secrets in our repo and we had also pinned the version of the
action to sha so there is no issue for us.

But the repo is gone now. I've instead matched this code to what we
are doing in the gitlab CI runner where we did not have such an
action anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,40 +24,29 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
+      - name: Calculate clone depth.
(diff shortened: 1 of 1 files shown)


willysr:
merged

willysr:
@aclemons the repo is now back online with the malicious commit removed
Do you want to keep this PR or back to original repo?

aclemons:
Let's use this PR since the code then matches what we do in `.gitlab-ci.yml`. It's fine to merge it during the normal weekly update if we are not doing one out of schedule.

PR #10131, opened 2025-03-16
Upgrade yadifa to 3.0.2

eleksir:
* Build system switched from autotools to cmake
* Patch required version of automake to match Slackware-15.0, as it works okay
* Change filesystem layout to match Slackware-15: /var/run -> /run, /var/yadifa -> /var/lib/yadifa
* Disable broken test
* Update default config
* Remove cmake actions that touch files in /etc/yadifs dir
* Remove weird quirk in dnscore.h


eleksir:
What it was in previous times - I do not know, but it gives me chance to look at yadifa' cmake scripts closer. They looks good (promising better word), but still contains weird mistakes(?) or leftovers from debugging-something process. So i have to patch this these moments.

eleksir:
And yes, it was something else, not openssl version.

willysr:
merged

PR #10136, opened 2025-03-16
systems/c-lcrypt: Upgraded for version 4.0.0

rizitis:
systems/c-lcrypt: Updated for version 4.0.0


rizitis:
There is also version 3.0.0 in slackbuilds/ready, I don't know how to do it in those cases...

willysr:
please rebase on top of my commit

PR #10153, opened 2025-03-18
python/python3-boto3: Update for 1.36.24

isaackwy:
This version bump is to synchronize with botocore 1.36.24 (the version at SlackBuilds.org).

I might go back to regular Slackware maintenance a few months later... I only got inspired from trying to fix the recent OpenSnitch build error.


willysr:
@sbo-bot: build python/python3-boto3

PR #10231, opened 2025-03-22
development/spyder: Clarify commenting on python3-spyder-kernels version restriction

isaackwy:
python3-spyder-kernels has versions later than 3.0.0.
I would like to slightly change the documentation to reflect this.

It's python3-spyder-kernels >= 3.0.0 (not only version 3.0.0 itself) that cannot run with spyder 5.4.0.


willysr:
merged

PR #10232, opened 2025-03-22
office/miktex: Updated for version 25.3

LinRs:


willysr:
@sbo-bot: build x86_64 office/miktex

willysr:
@LinRs this failed to build

LinRs:
> ! Change file entry did not match (file /tmp/SBo/repo.bgVpkS/build_miktex/miktex-25.3/Programs/TeXAndFriends/texjp/uptex/source/uptftopl.ch, l.53).

When building with MAKEFLAGS='-j24 -l13', every time it failed in a different location with "! Change file entry did not match".

It builds fine with MAKEFLAGS='-j24' or MAKEFLAGS='-l13'.

willysr:
@aclemons is it possible to configure such configuration?

aclemons:
> @aclemons is it possible to configure such configuration?

It is possible, but I don't think it makes sense in this case. There is obviously a race condition that upstream has not handled. I can make this fail on the build machine with `-j36` consistently with no load parameter, so just because it passed on one box with `-j24` does not mean it will work for all users who might want to build it from slackbuilds.org.

I think rather the script should handle it.

It seems freebsd has also run into this problem and disabled multiple jobs:

https://cgit.freebsd.org/ports/tree/print/miktex/Makefile#n66

Arch have also run into this problem and limit it to 2:

https://aur.archlinux.org/cgit/aur.git/commit/PKGBUILD?h=miktex&id=9c34a0e31e6ef158af8cdd5ec64b022745801b55
https://aur.archlinux.org/cgit/aur.git/commit/PKGBUILD?h=miktex&id=55d4d5ddf684d63e958d3bbf209997211097cb92

LinRs:
aclemons, you are right. The problem can be reproduced when building with -jN (N >= 30).
It would be safer to build with "-j1" due to race conditions causing intermittent build failures.

LinRs:
@sbo-bot: build x86_64 office/miktex

office/miktex/miktex.SlackBuild
```diff
@@ -114,7 +114,8 @@ cd build
     -DUSE_SYSTEM_URIPARSER=${SYSTEM:-OFF} \
     -DUSE_SYSTEM_ZZIP=${SYSTEM:-OFF} \
     -DCMAKE_BUILD_TYPE=Release ..
-  make
+  make || make clean && install -Dm644 $CWD/*.lzma -t ../build \
```
willysr, line 117: i suppose make -j1 should be enough?

willysr (commented):
use -j1 alone is enough ?

willysr:
@sbo-bot: build x86_64 office/miktex

willysr:
@LinRs please fix the checksum

willysr:
@sbo-bot: build x86_64 office/miktex

willysr:
merged

PR #10314, opened 2025-03-28
multimedia/MediathekView: Updated for version 14.2.0.

newHeiko:
Fails sbopkglint: The .so-files from the included JRE are not executeable and not stripped. I prefer not to change that to not mess with the JRE.


willysr:
@sbo-bot: build multimedia/MediathekView

willysr:
merged

PR #10335, opened 2025-03-31
system/OpenSnitch: Edit README

isaackwy:
I just found out from installing OpenSnitch on my Slackware system (in VirtualBox) that I need to make /etc/rc.d/rc.opensnitchd executable (so that OpenSnitch actually filters outgoing connections.)

I would like the README edited accordingly.


willysr:
yes, i normally set all rc.X 644 so owner need to set manually
merged

PR #10398, opened 2025-04-06
python/python3-pykeepass: Update for 4.1.1.post1

isaackwy:
The source pyproject.toml specifies pyotp as an additional requirement.
However, even without installing pyotp, python3-pykeepass still runs without errors.

pyotp is only an additional dependency for performing tests.
The following commit provides further context:
https://github.com/libkeepass/pykeepass/commit/e312269138eeccc73b8eab24cafe45698e5880a3


isaackwy:
@sbo-bot: build python/python3-pykeepass

willysr:
merged

PR #10471, opened 2025-04-14
system/trash-cli: Edit (reword) README

isaackwy:
I plan on updating python3-keyring to 25.6.0.

Following the Arch Linux repo, I also plan on adding (optional) shell completion support:
https://gitlab.archlinux.org/archlinux/packaging/packages/python-keyring

Therefore, I would like to edit the trash-cli README in preparation for adding that similar feature on python3-keyring.


willysr:
merged

PR #10472, opened 2025-04-14
academic/root: Updated for version 6.34.08

arabusov:
This MR removes the fixWriteFastArray.patch, because it is already included into the ROOT mainline.


willysr:
@sbo-bot: build academic/root

willysr:
merged

PR #10491, opened 2025-04-15
network/bitcoin: Updated for version 29.0.

nomnombtc:
Hello,

this update to 29.0 has some more changes than usually:

- remove qrencode detection, as it is included in a full install anyway. But autodetect zeromq if installed.
- switch to cmake build system ( but it needs  cmake >=3.22, so I had to add cmake-opt to requirements. )
- use make install/strip instead of using the extra block that did the stripping.
- updated the doc section to include more useful information ( and  a python script that is referenced in manpages. )


willysr:
@sbo-bot: build network/bitcoin

willysr:
merged

PR #10572, opened 2025-04-23
python/python3-grpcio: Actually remove the protobuf-reenable-private-headers patch

isaackwy:
I forgot to remove the patch for re-enabling private headers.
The patch was only necessary for protobuf 30.1.


willysr:
merged

PR #10576, opened 2025-04-23
gis/geopy: Edit README

isaackwy:
"pytz" is currently not available at SlackBuilds.org, as it has been split into python2-pytz and python3-pytz.
I would like to update the README in accordance with this.

EDIT: Also, remove trailing whitespace from README.


willysr:
merged

PR #10864, opened 2025-05-16
libraries/tepl: Fix building on Slackware-current

0xBOBF:
Detect the glib-2.0 version and set G_CONNECT_DEFAULT=0 only on Slackware-15.0 builds, so that tepl still builds on Slackware-current.


willysr:
merged

PR #10893, opened 2025-05-17
development/smartgit: Updated for version 24.1.3 and new maintainer

antonioleal:
development/smartgit: Updated for version 24.1.3 and new maintainer


antonioleal:
After checking with previous maintainer, I'm taking over his SlackBuilds.

PR #10902, opened 2025-05-18
python-aiostream: Downgrade version to 0.4.5

Tonus1:
Upgraded by mistake. It breaks the only SlackBuilds' software that depends on (vdirsyncer, that I maintain).
Didn't saw any complains, but sorry for that.
Let's note that the last version doesn't fix that.


willysr:
merged

PR #11007, opened 2025-05-29
Add to .gitignore: debian-binary

rizitis:


willysr:
if it's just 1 package that uses it, i'm not sure if it's worth adding

PR #11160, opened 2025-06-11
python/python3-plotly: Update for 6.1.2

isaackwy:
The updated pyproject.toml mentions hatch and jupyter-packaging as new build dependencies.
However, I don't think they're actually necessary.

The Arch Linux PKGBUILD of python3-plotly has the --skip-dependency-check build option activated:
https://gitlab.archlinux.org/archlinux/packaging/packages/python-plotly/-/blob/main/PKGBUILD


isaackwy:
@sbo-bot: build python/python3-plotly

willysr:
merged

PR #11274, opened 2025-06-22
user/schenkel/updates

gbschenkel:
Bumping podman and his deps to latest version.
Included netavark and aardvark-dns.

@willysr Since Podman uses netavark and aardvark-dns as default now, I need check where I need put a info about using the kernel module `ip_tables` if on netavark or podman.


willysr:
@gbschenkel netavark hasn't  been approved yet, so you need to submit it manually and you need to rebase from latest master since aardvark-dns has been pushed to master so no need to include it here

PR #11392, opened 2025-07-05
development/spyder: Update for 5.5.0

isaackwy:
I'm updating spyder only to the point that python3-lsp-server remains at version 1.3.0 (versions after that require python-packaging >= 22.0. Slackware has python-packaging 21.3.)

Also:
1. I'm fixing an error in which spyder does not run (incorrectly claims that backports_tarfile, jaraco.classes, jaraco.context and jaraco.functools are missing.)
2. I'm also fixing an incorrect claim of missing setuptools dependency (this happens after applying the replace-pkg_resources patch below.)


isaackwy:
@sbo-bot: build development/spyder

willysr:
merged

PR #11395, opened 2025-07-06
system/trash-cli: Fix installation of shell completions

isaackwy:
If python3-shtab is installed, but trash-cli is not, then trash-cli fails to install on my system.

The reason is:
Looking inside the for loop, the lines start with the following:
$CMD --print-completion ...

The lines should instead be the following:
./$CMD --print-completion ...

The loop currently runs commands like the following:
trash-empty --print-completion ...

The loop should instead run commands like the following:
./trash-empty --print-completion ...

The loop should run the files inside the /tmp/SBo/trash-cli-0.24.5.26 source folder, not the system files.


willysr:
merged

PR #11402, opened 2025-07-07
python/python3-pypandoc: Update for 1.15

isaackwy:
Also, edit README (indicate that pandoc can be installed instead of pandoc-bin).

For context, pandoc is compiled. pandoc-bin is a repackaged binary.


isaackwy:
@sbo-bot: build python/python3-pypandoc

willysr:
merged

PR #11427, opened 2025-07-08
python/python3-cookiecutter: Remove python3-jinja2-time dependency, add python3-arrow dependency

isaackwy:
1. python3-cookiecutter (since version 2.2.0) no longer depends python3-jinja2-time.
2. I am removing the python3-jinja2-time dependency. Since python3-jinja2-time itself depends on python3-arrow, I have to add back the python3-arrow dependency (python3-cookiecutter still depends on python3-arrow).
3. Also, bump copyright version year (but that's just an addition.)
4. I am not bumping the SlackBuild build number.


willysr:
merged

PR #11451, opened 2025-07-11
development/dart-sdk: Updated for version 3.7.3.

### LINKED ISSUE #11271: development/dart-sdk needs updating
I'm a maintainer of dart-sass. The developer decided to stop providing 32 bit binaries starting with version 1.89.1, therefore I had to switch to building that package from source. As the name implies, it depends on dart. However, dart-sdk on SBo is too old (2.10.4). Please update it to 3.7.3 (the last stable version to provide binaries for 32 bit systems). I checked it and it works on both 32 and 64 bit versions of Slackware 15.0.

There don't seem to be any other scripts depending on dart-sdk, so this update should be painless. No changes to the script seem to be required, just a version bump.

Also, the script doesn't make dart-sdk.sh executable. Arguably, if a user installs dart-sdk, they would want to source that file so dart is in their PATH. Google-go-lang script does that. 

I'm writing here because I never got a reply from @qunying after two weeks of waiting and was told to open an issue here on the mailing list.

fsLeg:
This resolves #11271

### DIFF
--- a/development/dart-sdk/dart-sdk.SlackBuild
+++ b/development/dart-sdk/dart-sdk.SlackBuild
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-# Slackware build script for <dart-sdk>
+# Slackware build script for dart-sdk
 
 # Copyright 2016-2020 Zhu Qun-Ying
+# Copyright 2025 Vladislav 'fsLeg' Borisov, Moscow, Russia
 # All rights reserved.
 #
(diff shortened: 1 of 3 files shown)


fsLeg:
@sbo-bot: build development/dart-sdk

fsLeg:
Aww, I'm not cool enough to trigger build check.

willysr:
@sbo-bot: build development/dart-sdk

willysr:
merged

PR #11466, opened 2025-07-12
development/bazel: Update for 8.3.1

isaackwy:
I previously kept bazel at 5.x.x because I planned on eventually uploading tensorflow.

I now don't plan on uploading tensorflow to SlackBuilds.org (at least until Slackware 15.1 or Slackware 16.0 gets released.)

Therefore, I would like to update bazel.

Also, I am not stripping the binary, as this will break other builds that depend on bazel.


isaackwy:
@sbo-bot: build development/bazel

willysr:
merged

PR #11504, opened 2025-07-17
python/pyfuse3: Updated deps

atelszewski:
Add Trio to the REQUIRES.
Trio is needed (at least) when pyfuse3 is used with borgfs.

NOTE:

I have the maintainer's permission to update the script.


willysr:
merged

PR #11660, opened 2025-07-30
system/pass-import: Update dependency from python3-zxcvbn-python to python3-zxcvbn

isaackwy:
python3-zxcvbn is the new version of python3-zxcvbn-python.

Also, bump build number in accordance with dependency change.

In addition, there is a file at `/usr/lib/password-store/extensions/import.bash`.
Must I change it to `/usr/lib${LIBDIRSUFFIX}/password-store/extensions/import.bash` (for example, `/usr/lib64/password-store/extensions/import.bash`?)


isaackwy:
@sbo-bot: build system/pass-import

willysr:
merged

PR #11687, opened 2025-08-01
python/python3-zxcvbn-python: Removed (no longer used)

isaackwy:
The README itself indicates that python3-zxcvbn-python is to be removed once it is no longer a dependency for anything.
I made pass-import depend on python3-zxcvbn rather than python3-zxcvbn-python.
(For context, python3-zxcvbn is the newer version of python3-zxcvbn-python.)

fourtysixandtwo provided me permission to make this change.


willysr:
merged

PR #11774, opened 2025-08-08
multimedia/totem: Patch meson build for graphics-update.

0xBOBF:
Added a patch from upstream which fixes building totem on Slackware-15.0 when the 'graphics-updates' from slackware-15.0's 'testing' branch are installed to the system.

Installations without the 'graphics-updates' installed are unaffected by the patch.


willysr:
merged

PR #11776, opened 2025-08-08
system/limine: Updated for version 9.6.0

antonioleal:
system/limine: Updated for version 9.6.0


antonioleal:
...And New Maintainer

willysr:
@sbo-bot: build system/limine

willysr:
think this is only for x86_64? i586 failed to build

willysr:
for i586, the ARCH should be set to --enable-uefi-ia32
see https://github.com/limine-bootloader/limine/blob/1cbc5ab01e9203acb166f838761ef633c1c6dc08/configure.ac#L215C22-L215C40

willysr:
you might need to use different variable as you need to set ARCH to i586 for SlackBuild

PR #11856, opened 2025-08-14
development/notepad++: Updated for version 8.8.5

antonioleal:
development/notepad++: Updated for version 8.8.5


willysr:
if the development cycle is releasing too fast, consider to push PR closer to weekend, so i don't have to remove old commits everytime you push new version every few days and keep the permission 644

willysr:
merged

PR #11911, opened 2025-08-18
Fix exec command , add douninst.sh

rizitis:


misc/balena-etcher-bin/douninst.sh
```diff
@@ -0,0 +1 @@
+rm -rf /opt/balena-etcher
```
willysr, line 1: i don't think this is the correct way to remove a directory
it should be removed via removepkg, not douninst.sh

willysr:
@sbo-bot: build misc/balena-etcher-bin

willysr:
merged

PR #11927, opened 2025-08-19
misc/bitwarden-desktop: updated for version 2025.3.1

maravtdm:
reverted to a glibc compliant version (2.33) for Slackware 15.0


willysr:
merged

PR #11957, opened 2025-08-22
desktop/AppStream: Update for 1.0.3

isaackwy:
Though Damien Perticone had allowed me to take over maintenance, I am not currently doing so.

The latest version of AppStream is at 1.0.6.
However, if I attempt to build AppStream 1.0.4, I receive the following error:
```
find: File system loop detected; ‘./tests/samples/compose/usr/share/locale/dir_loop/loop’ is part of the same file system loop as ‘./tests/samples/compose/usr/share/locale/dir_loop’.
chmod: cannot operate on dangling symlink './tests/samples/compose/usr/share/metainfo/org.example.nonexistent-badlink.metainfo.xml'
```

The symlink at `tests/samples/compose/usr/share/locale/dir_loop/loop` is causing this error.

I could replace `find -L . \` with `find . \` to successfully build AppStream >= 1.0.4; however, I don't want to break typical SlackBuild convention.

https://github.com/Ponce/slackbuilds/issues/316 provides further context for why I am making this update - flatpak 1.16.0 on Slackware-current depends on AppStream.

This update is not necessary for running flatpak in and of itself - however, I still have some small concerns that flatpak 1.16.0 requires AppStream >= 1.0.0.


isaackwy:
@sbo-bot: build desktop/AppStream

willysr:
merged

PR #11985, opened 2025-08-25
desktop/vertex-maia-icon-theme: Removed (orphan maintenance, and download link no longer works)

isaackwy:
The download link at GitHub no longer works.

I could change it to the website at GitLab; however, I am no longer maintaining vertex-maia-icon-theme. I don't know which commit to select for downloading vertex-maia-icon-theme.

For reference, here are the GitLab commits:
https://gitlab.manjaro.org/artwork/icon-themes/vertex-maia-icon-themes/-/commits/master?ref_type=HEADS

Also, for further context: I had uploaded the vertex-maia-icon-theme SlackBuild during the Slackware 14.2 era.


willysr:
merged

PR #12072, opened 2025-09-01
perl-PDF-API2: Updated for version 2.048

mab974:
no longer required by gscan2pdf
updated anyway but not tested.


willysr:
@sbo-bot: build perl/perl-PDF-API2

willysr:
merged

PR #12140, opened 2025-09-06
python/python-distutils-extra: Removed (replaced with python2 and python3 SlackBuilds)

isaackwy:
All the SlackBuilds that previously depended on python-distutils-extra:
I already had changed the dependency mentions.


willysr:
merged

PR #12142, opened 2025-09-06
system/depfinder: Updated script

antonioleal:
system/depfinder: Updated script

### DIFF
--- a/system/depfinder/README
+++ b/system/depfinder/README
@@ -1,4 +1,71 @@
 depfinder is a tool that finds the dependencies of Slackware packages.
 
+depfinder is much faster than requiredbuilder, a very nice script
+by Stefano Stabellini that I used in the past to find package
+dependencies. But requiredbuilder is written entirely in bash and I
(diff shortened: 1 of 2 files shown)


willysr:
@antonioleal there's a reason why i deleted the whole README and it's not going back again

antonioleal:
explain

willysr:
README is only used to display in the SBo website, so it should be brief and explain clearly about the software is all about, nothing else. That's why i cut most of the README content from your original submission and leave the important content only. In this PR, you want to restore the README back to the original README in upstream project repository. That's not what README is used for in our case.

antonioleal:
Ok Willy, understood. Thank you for the information.
Shall I post a new PR, with the previous README, or you'll do it when merging?

I did change the CXXFLAGS in this one..

willysr:
just push a new commit on top of this one should be ok

antonioleal:
Amended

willysr:
@sbo-bot: build system/depfinder

willysr:
merged

PR #12188, opened 2025-09-10
network/Electrum: Update README

nomnombtc:
This only updates the README file, no changes to the script itself.

informs the user that newer version would require python3.10, but that
there is also the option to use an AppImage provided on upstreams website.


willysr:
merged

PR #12261, opened 2025-09-16
misc/stardict: Patches for Slackware-Current build.

rizitis:


rizitis:
This is better solution to fix current build and avoid one more branch for [ponce](https://github.com/Ponce/slackbuilds/pull/326) repo...

willysr:
i will let @Ponce decide whether to include this in stable or in his repo for current

Ponce:
> This is better solution to fix current build and avoid one more branch for [ponce](https://github.com/Ponce/slackbuilds/pull/326) repo...

actually I don't agree with this: the commit overcomplicates the SlackBuild...
what happens when there's an update to stardict? the SlackBuild will change radically.
it's so much cleaner, IMHO, just to create a single patch and apply it in the unofficial repo.

rizitis:
I re open pull request as ponce suggested. so close this.

PR #12296, opened 2025-09-19
python/python3-pyinotify: If using Slackware current, make asyncore support optional

isaackwy:
This pull request is actually for Slackware current.

Nonetheless, I can still make it on this repo (for Slackware 15.0).

The context is that asyncore is removed in Python 3.12 (the version which is on Slackware current).
This patch addresses this issue (but will only activate if Python is at >= 3.12).


isaackwy:
@sbo-bot: build python/python3-pyinotify

willysr:
merged

PR #12297, opened 2025-09-19
system/lightdm: Add pam rules for fscrypt

isaackwy:
If I add the following rules to `/etc/pam.d/login`:
```
auth            optional        pam_fscrypt.so
session         optional        pam_fscrypt.so
```

And to `/etc/pam.d/passwd`:

```
password        optional        pam_fscrypt.so
```

Since I use fscrypt to encrypt /home, if I log in through the console (ex. tty1), then fscrypt automatically decrypts upon login.
However, this is not the case if I log in through lightdm.

I would like to correct this. I have added the option FSCRYPT=yes to pass onto the SlackBuild (if passed, add pam rules to allow automatic unlocking of fscrypt-encrypted directories.)

### DIFF
--- a/system/lightdm/README
+++ b/system/lightdm/README
@@ -14,3 +14,8 @@ Then, edit /etc/rc.d/rc.4 and add the following section:
 if [ -x /usr/bin/lightdm ]; then
   exec /usr/bin/lightdm
 fi
+
+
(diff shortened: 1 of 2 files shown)


isaackwy:
@sbo-bot: build system/lightdm

PR #12381, opened 2025-09-24
python/python3-zstandard: Updated for version 0.25.0

antonioleal:
python/python3-zstandard: Updated for version 0.25.0


antonioleal:
Sorry about sending this through PR, but SlackBuilds.org did not allow me to delete my previous submission.

This package is a dependency for the python3-nuika compiler that slipped past its first submission.

willysr:
this is a new submission, so it must go through normal submission first

PR #12423, opened 2025-09-26
multimedia/popcorntime: Download URL update.

fsLeg:
There was a hostile takeover of the upstream repo by previous developers who completely wiped the code and all uploaded artifacts. I'll mirror the sources for now, in the meantime I'll look into other forks.


aclemons (approved):
LGTM

PR #12454, opened 2025-09-28
removed python3-uv due to already existing uv-bin

antonioleal:
python3-uv was wrongly pushed to SBo.
The same package already existed under the name uv-bin.


willysr:
merged

PR #12525, opened 2025-10-04
python/pynacl: Remove from SlackBuilds.org (replaced with python3-pynacl)

isaackwy:
I have renamed pynacl to python3-pynacl.


willysr:
merged

PR #12672, opened 2025-10-15
development/jupyter-ipykernel: Update README

isaackwy:
jupyter-ipykernel >= 6.30.0 requires python-packaging >= 22.0.

Update README to indicate why jupyter-ipykernel should not be updated.


willysr:
merged

PR #12673, opened 2025-10-15
libraries/libminizip: Update README

isaackwy:
Update README.
Explain why I am keeping libminizip at version 1.2.13.


willysr:
merged

PR #12741, opened 2025-10-22
element-desktop: Updated for version 1.12.2.

fsLeg:


fsLeg:
@sbo-bot: build network/element-desktop

fsLeg:
@sbo-bot: build network/element-desktop-bin

fsLeg:
It seems this version changed something in the build process related to keytar. It built fine on my machine until I disconnected the internet after seeing the error here. I'll need to look deeper into what else needs to be vendored now. This PR might not be ready by this week's update.

fsLeg:
@sbo-bot: build x86_64 network/element-desktop

fsLeg:
What I found was the required file gets built fine, however node-gyp, when run under yarn, can't create/find some files if run offline (go figure), even though there's no such issue when calling it directly. The resulting package seems to behave normally if you just ignore the error, therefore I just added `|| true` at the offending `yarn install` step. If it wasn't safe to ignore, the script would've probably failed at a later point.

willysr:
merged

PR #12806, opened 2025-10-25
Update info files for electron-bin

rizitis:


willysr:
You need to make 3 separate commits for this

PR #13086, opened 2025-11-15
Update README (lxqt-build-tools, lxqt-build-tools-qt6)

isaackwy:
Improve readability.

1. Within each README, the phrase "This SlackBuild" is not clear enough. I would like to change that.
2. The lxqt-build-tools-qt6 README should start by describing lxqt-build-tools-qt6, not lxqt-build-tools. (I know that the original program name is lxqt-build-tools; however, the reader might get confused between lxqt-build-tools and lxqt-build-tools-qt6).


willysr:
merged

PR #13087, opened 2025-11-15
desktop/libqt6xdg: Update README and slack-desc

isaackwy:
Please read https://github.com/SlackBuildsOrg/slackbuilds/pull/13086 for further context.

The README should not start with "libqtxdg is a Qt implementation..."
libqtxdg is a separate SlackBuild - it is for qt5 support.


willysr:
merged

PR #13111, opened 2025-11-18
graphics/brlcad: updated for version 7.42.0

Ythogtha:
Crossing my fingers.
I've had a hard time building this, and decided to go full bundled libs, it seems to be working.
Also, it looks like having assimp installed breaks the build.
Let's try this...


Ythogtha:
@sbo-bot: build graphics/brlcad

Ythogtha:
Oh, this now requires complete network access.
Let me try and remove that requirement...

willysr:
@Ythogtha any update on this?

Ythogtha:
@willysr Still some work there. There are really an awful lot of possible dependencies if we decide to build everything as its bundled by brlcad (6Go of git pulls), so I'm trying to isolate what can be relied on by using system libs or SBos requirements, and what should be bundled.

It'll take some more time, so maybe we could close that PR for now.
I thought I had a solution because, as it is, with an internet connection, it builds, but we can do better...

willysr:
i'll put WIP on this so you can work on it when you have time

willysr:
i'll close this for now

PR #13258, opened 2025-11-28
academic/units: Fix compilation on -current

ArTourter:
No need to change build number as there is not need to recompile on -stable


ArTourter:
@sbo-bot: build academic/units

willysr:
mergd

PR #13270, opened 2025-11-30
development/nim: Updated for version 2.2.6.

nick8325:


nick8325:
Forgot to say, I checked that the reverse dependency `chawan` still works with the new nim.

willysr:
@sbo-bot: build development/nim

willysr:
merged

PR #13332, opened 2025-12-04
network/element-desktop: Updated for version 1.12.6.

fsLeg:


fsLeg:
@sbo-bot: build x86_64 network/element-desktop

fsLeg:
Hey, look, that thing I was asking about in email happened.

willysr:
yeah, i noticed because you put nodejs24-bin and yarn which uses older nodejs
i guess you need to make a new yarn that uses nodejs24 instead of the older nodejs

fsLeg:
That's a bit too much, because it'll be the same SlackBuild SBo already has, the only difference will be the REQUIRES field. Yarn already works as is with nodejs24-bin.

fsLeg:
I'll look into bundling yarn using corepack. Hopefully, it won't take longer than an hour or two tonight.

fsLeg:
@sbo-bot: build x86_64 network/element-desktop

willysr:
merged

PR #13338, opened 2025-12-04
development/llvm-opt: Use '-exec command {} +'

linrs:
- using '-exec command {} +' instead of '-exec command ;'. Thanks to Lockywolf.


willysr:
merged

PR #13399, opened 2025-12-11
libraries/libfm-qt: Remove lxqt-build-tools dependency

isaackwy:
The lxqt-build-tools dependency mention is redundant.

lxqt-menu-data already depends on lxqt-build-tools.


willysr:
merged

PR #13400, opened 2025-12-11
development/jupyter_client: Update README

isaackwy:
I cannot update jupyter_client to >= 8.7.0.

This commit is the reason why:
https://github.com/jupyter/jupyter_client/commit/bc5d2d5a6bc56e0d88245e8a89fc9e44c9e2d673

Upstream dropped Python 3.9 support since jupyter_client 8.7.0.


aclemons (approved):
LGTM

PR #13456, opened 2025-12-14
desktop/numix-gtk-theme: Update dep.

willysr:


willysr:
@sbo-bot: build x86_64 desktop/numix-gtk-theme

willysr:
maybe we need https://github.com/sass/listen or https://github.com/guard/listen ?

rizitis:
I think we need: 
https://github.com/guard/listen (which need:)
https://github.com/guard/rb-inotify (which need:)
https://github.com/ffi/ffi

willysr:
i'll just drop this from the repo as the maintainer is no longer active anyway

PR #13457, opened 2025-12-14
python/tinycss2: Removed (renamed to python3-tinycss2)

isaackwy:
I have renamed all the dependency mentions of tinycss2 last week.


willysr:
merged

PR #13458, opened 2025-12-14
python/python3-lsp-server: Restore default yapf >= 0.33.0 requirement

isaackwy:
Tonus already had upgraded yapf from 0.32.0 to 0.43.0 at the end of July of this year.
Therefore, I can bring back the default yapf version requirement (i.e. yapf >= 0.33.0).


willysr:
merged

PR #13603, opened 2025-12-22
development/spyder: Fix replace-pkg_resources patch

isaackwy:
When I attempted to run spyder, I received a runtime error.

The error is related to my patch.

Also, bump build.


willysr:
@sbo-bot: build development/spyder

willysr:
merged

PR #13604, opened 2025-12-22
development/jupyter-nbclient: Update README (why software version should not be updated)

isaackwy:
Upstream has dropped support for Python < 3.10.


willysr:
merged

PR #13607, opened 2025-12-22
multimedia/muse-sounds-manager: removal from SlackBuilds.org

antonioleal:
Unfortunately muse-sounds-manager is no longer usable with MuseScore.

muse-sounds manager is a binary distribution of samples, the latest release build with a more recent gcc preventing MuseScore from loading libMuseSamplerCoreLib.so

    11101:     /usr/lib64/libstdc++.so.6: error: version lookup error: version `GLIBCXX_3.4.30' not found (required by /home/user/.local/share/MuseSampler/lib/libMuseSamplerCoreLib.so) (fatal)
12:34:36.827 | ERROR | main_thread     | MuseSamplerResolver::init | Unable to load MuseSampler library: /home/user/.local/share/MuseSampler/lib/libMuseSamplerCoreLib.so

This by no means brakes MuseScore but Slackware users can now only used the standard application sounds.

Because there is no way to support older releases of muse-sound-manager, as only the latest version is made available, I ask for muse-sound-manager to be removed from SlackBuilds.org


willysr:
merged

PR #13671, opened 2025-12-28
libraries/plasma-wayland-protocols-opt: Fix HOMEPAGE

isaackwy:
https://invent.kde.org/libraries/plasma-wayland-protocols actually works.
https://invent.kde.org/frameworks/plasma-wayland-protocols leads to a sign-in page.

From here on out, I won't regularly update SlackBuilds.
In fact, I plan on quitting maintaining SlackBuilds after Christmas - but I found something I had to fix.
https://lists.slackbuilds.org/pipermail/slackbuilds-users/2025-December/030703.html

My future plans: when I get the time, I'll update the LXQt SlackBuilds series.


willysr:
merged

PR #13675, opened 2025-12-28
development/gh-cli: add gh-cli SlackBuild

r1w1s1:
Add a new SlackBuild for GitHub CLI (gh).

Tested on:
- Slackware 15.0
- Slackware-current


willysr:
new script should go via submission form

PR #13788, opened 2026-01-04
audio/mpd: Updated for version 0.23.17

cwilling:
This version update enables building with recent fmt update to 12.1.0

### DIFF
--- a/audio/mpd/mpd.SlackBuild
+++ b/audio/mpd/mpd.SlackBuild
@@ -106,6 +106,7 @@ cd build
     -Dpulse=$PULSE \
     -Dupnp=disabled \
     -Dmikmod=disabled \
+    -Dyajl=disabled \
     --buildtype=release \
     --infodir=/usr/info \
     --libdir=/usr/lib${LIBDIRSUFFIX} \


willysr:
@sbo-bot: build audio/mpd

aclemons:
Seems we patched out the yajl suffix in the package config:

https://github.com/SlackBuildsOrg/slackbuilds/blob/9196c7fd40663f9586e9f0265b45d5a9e49fd960/libraries/yajl/patches/0002-pkg-config-include-dir-should-not-have-the-yajl-suff.patch#L24

And it looks like mpd changed where it looks in this release:

https://github.com/MusicPlayerDaemon/MPD/commit/0aeda01ba6d22a8d9fc583faa67ffc6473869a43

might need to patch that back in

aclemons:
Or the 0.24.x branch wrote out the yajl dependency:

https://github.com/MusicPlayerDaemon/MPD/commit/248d96883a6cc69a3694d911983bcd88b7013b7d

cwilling:
What's with the hints that generate additional dependencies?

As far as  hard requirements go, mpd has only REQUIRES=fmt and fmt has REQUIRES="". Therefore an mpt build should only be dragging in fmt as a requirement - nothing else.

Yet from the failure report, many suggestions from the README are also dragged in as requirements. It is unclear how it is decided which ones end up being required - just that someone's hints are used. Now one of these arbitrarily chosen requirements is causing build failure for mpd. Somehow that doesn't seem right.

aclemons:
> What's with the hints that generate additional dependencies?
> 
> As far as hard requirements go, mpd has only REQUIRES=fmt and fmt has REQUIRES="". Therefore an mpt build should only be dragging in fmt as a requirement - nothing else.
> 
> Yet from the failure report, many suggestions from the README are also dragged in as requirements. It is unclear how it is decided which ones end up being required - just that someone's hints are used. Now one of these arbitrarily chosen requirements is causing build failure for mpd. Somehow that doesn't seem right.

We typically test with all the optional dependencies - those that are listed in the README as optional.

Since that's a manual process, they can sometimes get out of sync over time. If I'm doing the reviews, I'll usually try to check and sync the hints when I approve submissions. I synced the hints for mpd earlier to the latest from the readme on master. They'll be used when the build is next triggered.

yajl is a reverse dependency for about 20 packages - so for someone building with sbopkg on the target host etc, there is some chance that yajl will already be installed: https://slackbuilds.org/advsearch.php?q=yajl&stype=revdep, failing the build.

If you don't want to support yajl, we should remove it from the readme and force disable it if that is possible.

willysr:
> Or the 0.24.x branch wrote out the yajl dependency:
> 
> [MusicPlayerDaemon/MPD@248d968](https://github.com/MusicPlayerDaemon/MPD/commit/248d96883a6cc69a3694d911983bcd88b7013b7d)

0.24.x need newer GCC :(
i tried it before

willysr:
@cwilling perhaps can you try to replace yajl with nlohmann_json since it is also your script?

aclemons:
It might be simplest to just revert that commit I referenced:

https://github.com/MusicPlayerDaemon/MPD/commit/0aeda01ba6d22a8d9fc583faa67ffc6473869a43.patch

After unpacking the tarball in the slackbuild, apply with `patch -p1 -R` to revert and when the next slackware version rolls around, we can move to the 0.24.x or whatever branch is active at the time and forget about yajl.

I tried it here to check and it builds:

https://github.com/SlackBuildsOrg/slackbuilds/pull/13789

I don't use mpd, but you should be able to check locally if you have yajl installed and then build it.

cwilling:
@sbo-bot: build audio/mpd

willysr:
merged

PR #13804, opened 2026-01-04
Update wifitui to 0.10.0

r1w1s1:
Update wifitui to 0.10.0

This update also fixes embedded version reporting by using the SlackBuild
VERSION variable instead of git metadata, ensuring correct --version output
when building from release tarballs.


r1w1s1:
losing this PR and will re-submit shortly.

This branch was mistakenly created from my local fork containing OVERLAY history, so the PR includes unrelated commits. The actual wifitui 0.10.0 update itself is correct, but I want to re-open a clean PR based directly on upstream/master with only the relevant changes.

Apologies for the noise — a clean update PR will follow.

PR #13815, opened 2026-01-05
libraries/zstd-cmake: new maintainer.

linrs:
- new maintainer


willysr:
no need to bump BUILD if it's just a new maintainer

PR #13844, opened 2026-01-06
system/xbattmon: fix build failure with newer compilers; adopt maintainership

r1w1s1:
Fixes a build failure on newer toolchains caused by an Xlib predicate
prototype mismatch.

This package was previously orphaned; I am adopting maintainership.


r1w1s1:
@sbo-bot: build system/xbattmon

willysr:
merged

PR #13855, opened 2026-01-07
multimedia/spotify: Downgraded to version 1.2.74.477

RezaT4795:
Downgraded to version 1.2.74.477 due to libc and libm incompatibility.


willysr:
merged

PR #13867, opened 2026-01-08
system/android-file-transfer: Update README

isaackwy:
Though I had orphaned this SlackBuild:

Update README (command line option does not require qt6).
I have not disabled x86 downloads (since command line option does not require qt6, android-file-transfer may be built in both x86 and x86_64. However, the default option with qt6 gui will only allow x86_64 downloads).

Note: Next update cycle, I will disable x86 downloads for my lxqt SlackBuilds. This pull request just now is related to that.


willysr:
merged

PR #13907, opened 2026-01-10
games/PrismLauncher: Updated for version 10.0.1.

fsLeg:


fsLeg:
@sbo-bot: build games/PrismLauncher

willysr:
@fsLeg i moved to x86_64 only since qt6 is only for x86_64

fsLeg:
@willysr Technically, it's a shortcoming of qt6 from SBo as -current has 32 bit version of qt6 and there's nothing preventing PrismLauncher from building in x86.

willysr:
yes, but qt6 is current is not the identical version of what we had in SBo unfortunately

PR #13928, opened 2026-01-11
Update fcitx5 and related packages

ltlnx:
Update fcitx5 and related packages. Also move some packages into their correct categories.


ltlnx:
fcitx5 and libime need to be built first, as some packages depend on new versions of those packages

willysr:
@ltlnx you need to make separate commits for each script and use the format "category/script: Update for version XXX" as commit message
As for the category, if you wish to change them, you need to do it via submision form so the category in the database is updated as well. If you want to keep this PR, you need to keep the existing category

willysr:
@ltlnx ?

willysr:
i will close this one for now

PR #13933, opened 2026-01-12
Please disable lxqt SlackBuilds' x86 source downloads

isaackwy:
@willysr suggested this last week (SlackBuilds depending on qt6 do not build on x86.)


willysr:
merged

PR #14055, opened 2026-01-21
development/nodejs22: Updated for version 22.22.0.

fsLeg:


fsLeg:
@sbo-bot: build development/nodejs22

willysr:
@fsLeg are you sure this is the right path?
usr/doc/node -> nodejs22-22.22.0/node

fsLeg:
@willysr Yes. The actual directory is /usr/doc/nodejs22-22.22.0/node but some things may (or may not?) expect files at /usr/doc/node. I'm not sure if /usr/doc is supposed to have files that aren't in $PRGNAM-$VERSION directories, hence a symlink. Although I have no idea what those scripts are and what expects them, maybe they can even be removed completely, but NodeJS installs them, so there must be a reason.

willysr:
merged

PR #14139, opened 2026-01-27
system/lightdm-settings: Update for 2.1.1

isaackwy:
This is the last update I will make for this weekly cycle.

I was getting my Linux Mint SlackBuilds updated (so far, not including xreader).
These SlackBuilds are ones I had used for my desktop (and still would use if I get back to Slackware).


isaackwy:
@sbo-bot: build system/lightdm-settings

willysr:
merged

PR #14205, opened 2026-01-30
academic/units: Updated for version 2.25.

ArTourter:


ArTourter:
@sbo-bot: build academic/units

PR #14211, opened 2026-01-31
python/python2-websocket-client: Orphaned.

maramon:


willysr:
@maramon this is required by python2-websocket-client
maybe @audriusk want to take over?

audriusk:
I'm not maintaining my SlackBuilds any longer, but regarding python2-docker, I'd say it should be removed completely (together with python2-docker-pycreds and python2-dockerpty). The version on SBo is from 2018, very likely not functional with the latest docker. python3-docker is the latest version and what should be used instead by everybody.

willysr:
ok, i will just remove it together

thanks @audriusk

PR #14267, opened 2026-02-04
office/pdfarranger: Reverse to version 10.0.1 because of pikedpf-5.3.2 on Slackware-15.0

brobr:
This is the last version that can use the dependency pikepdf < 6


willysr:
merged

PR #14273, opened 2026-02-04
development/kForth-64: Updated for version 0.6.1

antonioleal:
development/kForth-64: Updated for version 0.6.1


aclemons (approved):
LGTM

PR #14359, opened 2026-02-12
system/virtiofsd: Updated for version 1.13.3; new maintainer

duncan-roe:
This is the latest version from gitlab. mount_root.patch is no longer needed. Matteo said I could take maintainership in the last para of https://lists.slackbuilds.org/pipermail/slackbuilds-users/2025-November/030668.html


duncan-roe:
Possibly virtiofsd could be listed as an optional dependency of something - maybe virt-manager?

PR #14362, opened 2026-02-12
libraries/grpc: Update for 1.78.0

isaackwy:
Also, I am testing to see whether Bear and sysdig compile successfully.


willysr:
merged

PR #14391, opened 2026-02-14
network/munge: Update version to 0.5.18, security fix 

emman:
Update to version 0.5.18, fixes security bug: 
https://github.com/dun/munge/security/advisories/GHSA-r9cr-jf4v-75gh
All users should update to this version.


willysr:
merged

PR #14393, opened 2026-02-15
office/LibreOffice: New -srcs file & MD5SUM

cwilling:
recheck i586 build


cwilling:
Marking 32bit version as UNSUPPORTED (in line with office/libreoffice)

PR #14408, opened 2026-02-17
desktop/slstatus: add backported patches from upstream

r1w1s1:
Add patches/ directory with four commits backported from upstream
(https://git.suckless.org/slstatus):

- 8723e8b: more concise memory calculation on Linux (lscanf, /proc/meminfo)
- 6fa36ba: volume: avoid NULL dereference in onval() on sndio device switch
- 4f61bbb: wifi: always retry ioctl for ifindex
- 98769df: fix buffer overflow in battery state parsing (state[12] -> state[13])

Patches apply cleanly against slstatus-1.1 source.


r1w1s1:
@sbo-bot: build desktop/slstatus

PR #14465, opened 2026-02-24
gis/eccodes: Updated for version 2.46.0

ArTourter:
- using source from github
- compiling with OpenMP threaed rather pthreads


ArTourter:
well looks like the github source is not the correct one to use.
reverting back to the one released on confluence site

willysr:
merged

PR #14220, opened 2026-02-01
academic/ncview: Updated for version 2.1.11.

willysr:

PR #14510, opened 2026-02-27
libraries/netcdf: Updated for version 4.10.0.

ArTourter:
- change build to cmake

### DIFF
--- a/academic/ncview/ncview.SlackBuild
+++ b/academic/ncview/ncview.SlackBuild
@@ -75,6 +75,7 @@ find -L . \
  \( -perm 666 -o -perm 664 -o -perm 640 -o -perm 600 -o -perm 444 \
   -o -perm 440 -o -perm 400 \) -exec chmod 644 {} \;
 
+CC=$(nc-config --cc) \
 CFLAGS="$SLKCFLAGS" \
 ./configure \
   --prefix=/usr \


aclemons:
@sbo-bot: build academic/grace

aclemons:
@sbo-bot: build academic/nco

willysr:
@sbo-bot: build x86_64 academic/ncview

aclemons:
> @aclemons can you take a look on this? configure: error: no acceptable C compiler found in $PATH it was working fine on #14220

checking

aclemons:
@sbo-bot: build academic/ncview

aclemons:
I manually did this now - it's the netcdf upgrade in this PR which is breaking it:

```sh
root@4bb62e3976e1:/tmp/SBo/ncview-2.1.11# export SLKCFLAGS="-O2 -fPIC"
root@4bb62e3976e1:/tmp/SBo/ncview-2.1.11# export  LIBDIRSUFFIX="64"
root@4bb62e3976e1:/tmp/SBo/ncview-2.1.11# ./configure --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc --localstatedir=/var --mandir=/usr/man --docdir=/usr/doc/ncview-2.1.11 --with-nc-config=/usr/bin/nc-config --build=x86_64-slackware-linux
user specified nc-config is /usr/bin/nc-config
checking for nc-config... yes
Netcdf library version: netCDF 4.10.0
Netcdf library has version 4 interface present: yes
Netcdf library was compiled with C compiler: /usr/bin/cc
checking for /usr/bin/cc... no
configure: error: in `/tmp/SBo/ncview-2.1.11':
configure: error: no acceptable C compiler found in $PATH
See `config.log' for more details
```

This has changed:

New:
```
# nc-config --version
netCDF 4.10.0
# nc-config --cc
/usr/bin/cc
```

old:
```
# nc-config --version
netCDF 4.9.3
# nc-config --cc
gcc
```

It seems the configure script for ncview then looks for the literal full string "/usr/bin/cc" on `PATH` which of course does not exist.

We can patch the nc-config output here to make it return just `cc` or we can work around it in the ncview slackbuild - it does however look like the ncview maintainer is MIA.

I think the new behaviour is valid, it's just breaking ncview which is not prepared for it.

ArTourter:
This is really weird because running configure for ncview with
`CC=$(nc-config --cc) ./configure` actually works despite being the same. Hard-coding to /usr/bin/cc also works but it checks whether it matches the output of `nc-config --cc` anyway, so if that changes again, the configure will stop working again.

I am trying to find the what actually fails in the configure script of ncview as it is most likely a bug but I have never really played with autoconf at that level. So setting the CC variable might be the easiest way.

@aclemons Do you want me to add the commit to fix that?

aclemons:
> I am trying to find the what actually fails in the configure script of ncview as it is most likely a bug but I have never really played with autoconf at that level. So setting the CC variable might be the easiest way.
> 
> @aclemons Do you want me to add the commit to fix that?

Yes, if just setting CC in the ncview build is enough, I think that is the most practical solution. All the other things that need netcdf built without issue.

willysr:
merged

PR #14537, opened 2026-02-28
games/scummvm: Removed gold linker

wkdjgf534:
Removed gold linker. 

This linker brings problem to compile binary file on Slackware current.


willysr:
merged

PR #14541, opened 2026-03-01
python/python3-regex: Update for 2026.1.15

isaackwy:
There are newer versions of python3-regex.

However, this is the last version that supports Python 3.9.

I am not sure if later versions (ex. 2026.2.19) work on Slackware 15.0. Because of this, I have not changed the README to reflect the (actual?) version limit.


isaackwy:
@sbo-bot: build python/python3-regex

PR #14558, opened 2026-03-01
development/kForth-64: Fix script

antonioleal:
development/kForth-64: Fix script


antonioleal:
@sbo-bot: build development/kForth-64

PR #14573, opened 2026-03-01
games/gargoyle: updated for version 2026.1.1

Ythogtha:
re-pushed, forgot to add qt6 dependency.
Which makes it unsupported on 32bits.


willysr:
merged

PR #14589, opened 2026-03-02
system/incus-ui: Updated for version 20260226_944b3601

cwilling:
Build test for new version


cwilling:
Thanks, Just noticed that too. Will redo.

PR #14686, opened 2026-03-07
python/python3-pyrsistent: Update README

isaackwy:
python3-pyrsistent 0.21.0 has dropped support for Python 3.9 (it now requires Python >= 3.10.)


willysr:
merged

PR #14691, opened 2026-03-07
system/vim-airline: Orphan maintenance of vim-airline

isaackwy:
I don't plan on maintaining and using vim-airline anymore.
(I instead use the lightline plugin.)


willysr:
merged

PR #14800, opened 2026-03-15
network/signal-desktop: Reverted to version 8.1.0.

ArTourter:
The 8.2.0, and above, packages are compiled against a newer glibc.


willysr:
merged

PR #14928, opened 2026-03-25
academic/bowtie2: Update to version 2.5.5

brobr:


brobr:
Thanks Willy, but the cmake differences could be a red herring. The bowtie2 script uses `make` and the error referred to the Makefile.

And also the `make` versions differ: 15.0 has `make-4.3`, current `make-4.4.1`. The error referred to line 400 in the Makefile and one of the many changes in the Makefile for 2.5.5 vs 2.5.4 relates to the error. I am going to ask upstream what is best especially because the Manual for Bowtie2 mentions `make` as the means to compile the program from source, not `cmake` (actually both cmake-versions are much higher than 3.5/3.1; it did compile with `cmake-opt`).

##### Please ignore old question: 

    What is the best slackbuilds way to adapt the cmake template to have `cmake` calling `/opt/cmake-opt/cmake` only on 15.0 but not on current? Could you give me any pointers to examples? (Sorry,  I am not familiar with developing/building C programs; just took over the original script from Petar in 2018.). 

Searching for reverse depencies on slackbuilds.org gave me some examples (like https://slackbuilds.org/slackbuilds/15.0/gis/qgis/qgis.SlackBuild)

PS: I deleted previous message

PR #14975, opened 2026-03-28
python/jsonpatch: Removed (renamed to python3-jsonpatch)

isaackwy:
I already have renamed jsonpatch to python3-jsonpatch.


willysr:
merged

PR #14976, opened 2026-03-28
python/jsonpointer: Removed (renamed to python3-jsonpointer)

isaackwy:
I already have renamed jsonpointer to python3-jsonpointer.

I previously said that this is my last pull request to SlackBuilds.org until a new version of Slackware gets released (or at least is in alpha.)

However, I would like to get bemenu updated, but in another pull request.
That will be my actual last SlackBuild update/contribution (until Slackware 15.1/16/etc. is about to get released.)


willysr:
merged

PR #14983, opened 2026-03-30
desktop/bemenu: Update for 0.6.23 (+new maintainer)

isaackwy:
I had received permission from adcdam to take over maintenance.

Also, this is my actual last SlackBuilds update until a new version of Slackware is at least in alpha.

I announced that I will no longer maintain SlackBuilds last December... but I kept going until just now.

The reason for taking over maintenance is that I want to use bemenu when I return to Slackware.


isaackwy:
@sbo-bot: build desktop/bemenu

willysr:
merged

PR #15030, opened 2026-03-31
office/gospel-pdf-viewer:Removal request;use pdf-bunny

lecramyajiv:
upstream has changed the name of gospel-pdf-viewer to pdf-bunny.
pdf-bunny is in ready queue. please remove the gospel-pdf-viewer.


willysr:
merged

PR #15167, opened 2026-04-11
Gcc14 fixes

cwilling:
For fixes only, use [development/gcc14: Symlink & doc fixes](https://github.com/SlackBuildsOrg/slackbuilds/commit/bc28edda129307e0b3347b195a877973ef182727)


cwilling:
Willy already fixed by hand.

PR #15173, opened 2026-04-12
system/lxqt-config: Edit README (grammar fix)

isaackwy:
Here's an analogy regarding this README change:
Blue mixed with red become purple.

Not:
Blue mixed with red becomes purple.


willysr:
merged

PR #15195, opened 2026-04-13
python3-boto3 will no longer support Python 3.9 from 29th April 2026

antonioleal:
python3-boto3 will no longer support Python 3.9 from 29th April 2026.

Please remove SlackBuild python3-boto3.


willysr:
no need to drop boto3 since the current version will still work with 3.9, just no more updates until next slackware repo is ready

PR #15343, opened 2026-04-22
desktop/dwmblocks: Update for 20260418 (+edit README)

isaackwy:
The following section in the README is too unclear:
```
For better window manager stability, replace pulseaudio by installing a
newer version of pipewire (>= 0.3.52; for context, Slackware 15.0
provides pipewire 0.3.44) and wireplumber (>= 0.4.11, at SBo).
```

Back when I was using dwm with dwmblocks, I would use pipewire instead of pulseaudio as the audio backend.

I would also use newer versions of pipewire and wireplumber (so that I can change the volume by scrolling the dwmblocks audio module up and down.) This was actually the reason to upgrade pipewire to a version >= 0.3.52 and wireplumber to a version >= 0.4.11 - it wasn't because dwmblocks wouldn't work without the pipewire and wireplumber version upgrades.

Also, what does it mean to "replace pulseaudio"?


isaackwy:
@sbo-bot: build desktop/dwmblocks

willysr:
merged

PR #15385, opened 2026-04-26
python/python3-pypandoc: Update for 1.17 (+orphan maintenance)

isaackwy:
I am not using python3-pypandoc anymore.
This is my last update for python3-pypandoc.

I uploaded python3-pypandoc because it was once a dependency for pass-import (in particular version 3.3), but this is no longer the case.

Also:
This build works on 32-bit as well (it's just that the listed dependency is pandoc-bin, which itself does not support 32-bit.)


isaackwy:
@sbo-bot: build python/python3-pypandoc

willysr:
merged

PR #15563, opened 2026-05-09
python/python3-gensim: Fix build

isaackwy:
My goal is to get python3-scipy updated to 1.13.1. I am checking its dependencies.
I found out that python3-gensim currently does not build.

The build error has nothing to do with python3-scipy, but rather is related to python3-numpy.

Personally, I don't like this approach where the user has to install both python3-numpy and python3-numpy-legacy on the same system. Furthermore, I don't like to have to install python3-numpy-legacy at all.

Regardless, this is what I have to do to fix the build.

### DIFF
--- a/python/python3-gensim/python3-gensim.SlackBuild
+++ b/python/python3-gensim/python3-gensim.SlackBuild
@@ -2,7 +2,7 @@
 
 # Slackware build script for python3-gensim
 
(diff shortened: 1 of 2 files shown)


isaackwy:
@sbo-bot: build python/python3-gensim

aclemons (approved):
LGTM

PR #15567, opened 2026-05-09
libraries/Botan3: Removed (use Botan).

fsLeg:
Now that Botan was updated, there's no longer a need for Botan3.


willysr:
merged

PR #15652, opened 2026-05-14
system/hplip-plugin: reverted to version 3.20.5

Ythogtha:
hplip-plugins needs to be the same version as hplip.
The previous update was made for -current and not 15.0, as it bumped the version to -current version of hplip.

Since SBo follows stable, I have to revert it to the same version as -15.0.

I do have a printer that requires hplip-plugins, and with a -15.0 box, it fails using the updated hplip-plugins, and works with the proper matching version, so I am sure of what I am saying.

To be noted: the person who sent the update through the website form usurped my name and email, I never sent it, and never wanted to send it. I never use the form for updates.


Ythogtha:
@sbo-bot: build system/hplip-plugin

Ythogtha:
@sbo-bot: build system/hplip-plugin

willysr:
merged

PR #15684, opened 2026-05-15
python/python3-babel: fix when more than one cldr files are in the script directory

Ythogtha:
This is really minor, and will break the build if for example someone has an older cldr file in the SlackBuild directory.
The line stating `ln -s $CWD/cldr-common-*.zip cldr/cldr-common-$CLDRVER.zip` will break.
I'm being completely explicit about what is expected to be linked.
It changes nothing for the build itself, so I'm not bumping the build number.


Ythogtha:
@sbo-bot: build python/python3-babel

willysr:
merged

PR #15699, opened 2026-05-16
system/slackpkg-update-notify: Updated for version 1.0

rizitis:


rizitis:
@sbo-bot: build system/slackpkg-update-notify

PR #15892, opened 2026-05-27
libraries/vmaf: Slackbuild script update

lecramyajiv:
Enable building the vmaf binary. Also make it a
shared library only.


lecramyajiv:
@sbo-bot: build libraries/vmaf

willysr:
merged

PR #15928, opened 2026-05-29
misc/ollama: Updated for version 0.24.0.

linrs:
- new option VULKAN: building ggml-vulkan if you have vulkan-sdk >= 1.4.

- new option LIB64: installing ollama libs to /usr/lib$LIBDIRSUFFIX/ollama
  instead of hardcoded /usr/lib/ollama. Thanks to Christoph Willing for the patch.

### DIFF
--- a/misc/ollama/README
+++ b/misc/ollama/README
@@ -15,6 +15,12 @@ ROCM=ON: building with ROCm, default is ROCM=OFF, build and runtime
 dependencies:
   development/rocmtoolkit_7
 
+LIB64=OFF: supporting LIBDIRSUFFIX, for x86_64, ollama libs will be
+installed to /usr/lib64/ollama, default is LIB64=OFF.
+
+VULKAN=OFF: building without vulkan, default is VULKAN=OFF. If you
+have vulkan-sdk >= 1.4, enable it with VULKAN=ON.
+
 Building ollama requires network and google-go-lang.
 
 To verify the installation:
--- a/misc/ollama/ollama.SlackBuild
+++ b/misc/ollama/ollama.SlackBuild
@@ -26,7 +26,7 @@ cd $(dirname $0) ; CWD=$(pwd)
 
 PRGNAM=ollama
 VERSION=${VERSION:-0.24.0}
-BUILD=${BUILD:-1}
+BUILD=${BUILD:-2}
 TAG=${TAG:-_SBo}
 PKGTYPE=${PKGTYPE:-tgz}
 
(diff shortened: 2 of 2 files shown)


linrs:
@sbo-bot: build misc/ollama

cwilling:
@sbo-bot:  build misc/ollama

cwilling:
I don't understand the changes for x86_86 depending on the line:
`if [[ "${LIB64:-OFF}" = "ON" && "$ARCH" = "x86_64" ]]; then`
It means the default is to put 64bit libraries in /usr/lib - not /usr/lib64. The default should be to put 64bit libraries in /usr/lib64. This choice defaults to doing the wrong thing.

See sbo-bot's sbopkglint output from above:
```
Running test: 20-arch...
___ note: 64-bit x86_64 package contains /usr/lib; this may or may not be a problem.
```

Although upstream wrongly (for Slackware) puts 64bit libraries in /usr/lib, we should just put them into /usr/lib64 rather than making the user explicitly choose to do so.

Also, depending on ARCH = x86_64 is unnecessary since the following changes all use $LIBDIRSUFFIX i.e. they only have an effect if on x86_64

fsLeg:
If the path is hardcoded, wouldn't it make more sense to just use `/usr/lib$LIBDIRSUFFIX/ollama` as a prefix? Patching sources is more invasive, and there's a risk of missing something which could break the application, especially in the light of a major change that's coming in 0.30. Take a look at how I did it, maybe it'll inspire you: https://github.com/fsLeg/sb/blob/master/ollama/ollama.SlackBuild

cwilling:
Changing the prefix isn't enough in this case. At runtime, hardware acceleration (if available) doesn't work because although the shared libraries would have been installed in /usr/lib64, the cuda shared library can't be found (only /usr/lib is searched). That's what the  second two sed commands fix.

fsLeg:
@cwilling Incorrect. The hardcoded library path is `../lib/ollama` relative to `ollama` binary: https://github.com/ollama/ollama/blob/main/docs/development.md#library-detection So if PREFIX is `/usr`, the binary is installed as `/usr/bin/ollama` and the libraries are searched inside `/usr/lib/ollama/`. However, if PREFIX is `/usr/lib64/ollama`, then the binary is installed as `/usr/lib64/ollama/bin/ollama` and the libraries are searched inside `/usr/lib64/ollama/lib/ollama/`.

Also, I've been building it like this for over a month, and CUDA libraries are properly detected.

cwilling:
Frankly, I'm not enthusiastic about installing and searching for shared libraries (depending on your PREFIX) in either `/usr/lib/ollama/` or `/usr/lib64/ollama/lib/ollama` (of course `/usr/lib/ollama` is fine for 32bit libraries).

Also, while `/usr/bin/ollama`  is fine for a binary, the choice of `/usr/lib64/ollama/bin/ollama` for the 64bit binary is also non-standard.

Any locations can be made to work and I'm glad those locations work for you but they're unnecessarily unusual for me - especially when standard locations are achievable so easily.

fsLeg:
I wouldn't call hunting for hardcoded paths in the source and patching a potential moving target easy. And it seems pretty normal for programs that don't play by the rules to be put inside a separate directory inside `/usr/lib64` or `/usr/libexec` with a symlink or a wrapper script inside `/usr/bin`. Slackware is an outlier here, having a `lib64` directory instead of `lib` because history, the fact that this hardcoded path works virtually everywhere but Slackware is a testament to that. Is it a good practice? Of course not. But the program is here, the program is good and it's a package maintainer's job to make sure the program works in their distribution. How to do it is up to them, I just offered a working option.

willysr:
All we have to make sure is that the application is working well, even if it's not placed under standard location
Some upstream project do hard-code it to look for certain places and it's a burden for us to look for each applications. As long as the maintainer confirms it's working for him, that's fine by us.
I'm going to merge this PR since it's working already

PR #15935, opened 2026-05-29
libraries/grpc: Update README, mark as x86_64 only

isaackwy:
I am not updating grpc. It is meant to synchronize with the python3-grpcio version (they're both the same program but split into 2 different SlackBuilds.)

Also, I am running sbo-bot (to see if @willysr's fix actually worked for x86 systems.)

### DIFF
--- a/libraries/grpc/README
+++ b/libraries/grpc/README
@@ -5,3 +5,7 @@ of connected systems.
 
 This SlackBuild builds gRPC in C++. python3-grpcio contains the Python 3
 build of gRPC.
+
+grpc is kept back at 1.80.0 on Slackware 15.0. The program version is
+meant to synchronize with python3-grpcio (which cannot be updated due to
+requiring Python >= 3.10).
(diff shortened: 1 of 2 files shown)


isaackwy:
@sbo-bot: build libraries/grpc

willysr:
@isaackwy it still failed to build on i586

isaackwy:
> @isaackwy it still failed to build on i586

See what else you can do to get grpc 1.80.0 to build on i586.
If all fails, I will revert grpc to 1.78.0 (and then orphan the SlackBuild).

willysr:
another possibility is to mark it as x86_64 only
only sysdig depends on grpc anyway
what do you think @isaackwy ?

isaackwy:
> another possibility is to mark it as x86_64 only only sysdig depends on grpc anyway what do you think @isaackwy ?

Sure, I'll rebase this pull request shortly (in about 10 minutes or so) with the x86_64 only marking.

I'll make that the solution for now.
I might discuss this on the Slackware linuxquestions forum (a counterargument for x86_64 only is that there are packages in the future that might be good to have 32-bit builds (and therefore 32-bit grpc as a dependency.)

isaackwy:
@sbo-bot: build libraries/grpc

willysr:
merged

PR #15947, opened 2026-05-31
perl/perl-http-daemon: Switch to Build.PL.

willysr:


willysr:
@sbo-bot: build perl/perl-http-daemon

PR #16002, opened 2026-06-04
development/claude-code: Upgraded for version 2.1.161

NicholasBHubbard:
claude-code now ships as a [Bun-compiled](https://bun.com/docs/bundler/executables) single-file executable. Stripping discards the embedded payload, leaving it to run as the bare Bun runtime instead of Claude Code, so it cannot be stripped without breaking it. This is why sbopkglint's 20-arch test fails.


willysr:
@sbo-bot: build development/claude-code

willysr:
merged

PR #16062, opened 2026-06-07
multimedia/ytdlp-gui: Orphaned

rizitis:


rizitis:
This package not run properly in slackware-15.0 the best is to remove it as there are alternatives for the same job in SBo.

willysr:
removed on my branch

PR #16063, opened 2026-06-08
Change Yanes Checcacci Balod's email address

isaackwy:
Regarding my taking over of naps2:
My email failed to deliver to yanes@pobox.com.
However, it delivered to yanes@infolink.com.br.

Yanes had approved of this change.

I think the problem is that pobox is no longer a separate email service - it was acquired by Fastmail in 2015.


cwilling:
Merged

PR #16132, opened 2026-06-14
system/slapt-get: Fix build with meson.

willysr:


willysr:
@sbo-bot: build system/slapt-get

willysr:
merged

PR #16137, opened 2026-06-14
multimedia/gst-plugins-ugly: Fix build.

willysr:


willysr:
@sbo-bot: build multimedia/gst-plugins-ugly

PR #16172, opened 2026-06-16
system/lightdm: Support Wayland Sessions.

willysr:
@isaackwy can you review this PR?
This patch added wayland sessions support into lightdm

### DIFF
--- a/system/lightdm/default-config.patch
+++ b/system/lightdm/default-config.patch
@@ -13,7 +13,7 @@
  #backup-logs=true
  #dbus-service=true
  
-+sessions-directory=/usr/share/xsessions
++sessions-directory=/usr/share/xsessions:/usr/share/wayland-sessions
 +
  #
  # Seat configuration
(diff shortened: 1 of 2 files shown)


isaackwy:
I'll allow this.

Have not checked whether the patch works, but I'll give some context.

The default-config.patch file was derived from the following source:
https://slackware.uk/slint/x86_64/slint-15.0/source/lightdm/default-config.patch

That is, from slint's source builds.
The slint repo originally had `+sessions-directory=/usr/share/xsessions` and on my SlackBuild, you want to change it to `+sessions-directory=/usr/share/xsessions:/usr/share/wayland-sessions`

willysr:
yes, it supports multiple directories and this way, it will list wayland sessions. Otherwise, it will list X11 only

isaackwy:
Please make this change then.

willysr:
thanks
merged
i incorporate it into my CSB project as well

PR #16203, opened 2026-06-18
development/jupyterlab: Update README

isaackwy:
jupyterlab 4.5.9 and 4.6.0 have been released.

I am not updating jupyterlab to either version.
1. jupyterlab 4.6.0 requires Python >= 3.10 (which is not available on Slackware 15.0)
2. While I can update jupyterlab to 4.5.9, jupyter-notebook does not have a corresponding version at 7.5.8. There are jupyter-notebook versions 7.5.7 and 7.6.0.


cwilling:
Merged

PR #16222, opened 2026-06-18
network/mullvad-browser: Updated for version 15.0.16

antonioleal:
network/mullvad-browser: Updated for version 15.0.16


antonioleal:
@sbo-bot: build network/mullvad-browser

cwilling:
@antonioleal 
While processing this update I noticed that the upstream package is made so that it unpacks into `/usr/lib` - not `/usr/lib64` which would be much nicer for Slackware. I guess upstream is packaging for Ubuntu etc.

As a test, I locally
- moved the /usr/lib/mullvad-browser directory into /usr/lib64
- changed the mullvad-browser symlink in /usr/bin to point to new location in /usr/lib64/mullvad-browser directory
- changed the Exec target in /usr/share/applications/mullvad-browser.desktop to point to the new location

Quick testing found no problems with the new locations - it works really well.

It would be great if you could consider something like this for your next update.

cwilling:
Merged

PR #16250, opened 2026-06-20
system/monit: Updated for version 6.0.0.

willysr:


willysr:
@sbo-bot: build system/monit

PR #16277, opened 2026-06-23
Stellarium

willysr:


willysr:
@sbo-bot: build academic/stellarium

willysr:
@aclemons any idea why it can't find even though i added the commit before Stellarium update?

cwilling:
@willysr  @aclemons 
FWIW I can build this at home in a clean container with all deps preinstalled

willysr:
ok, you can push it to your branch, but i still need to find out why rebase is not working

cwilling:
> ok, you can push it to your branch, but i still need to find out why rebase is not working

I found the problem.

When home build succeeded, I had preinstalled all requirements.

The build here failed because stellarium.info REQUIRED contained QXslx but should have been QXlsx. Therefore even though script for QXslx existed in the rebase, it was never built so that stellarium could use it. From build log above:
```
WARNING: academic/stellarium: Dependency QXslx does not exist and has been ignored
```
I fixed now in my branch and pushed it to SBo.

PR #16369, opened 2026-07-01
network/cisco-secure-client-vpn: Updated for version 5.1.18.314.

RezaT4795:


RezaT4795:
@sbo-bot: build network/cisco-secure-client-vpn

cwilling:
@RezaT4795 
The final package contains several .rpm files in the root directory. If they are actually needed, could you store them elsewhere please?

Also, after removing systemd directory, nothing remains in /usr/lib so that could also be removed i.e. `rm -rf "$PKG/usr/lib"`, not just `rm -rf "$PKG/usr/lib/systemd"`

RezaT4795:
> @RezaT4795 
> 
> The final package contains several .rpm files in the root directory. If they are actually needed, could you store them elsewhere please?
> 
> 
> 
> Also, after removing systemd directory, nothing remains in /usr/lib so that could also be removed i.e. `rm -rf "$PKG/usr/lib"`, not just `rm -rf "$PKG/usr/lib/systemd"`

No, they are other packages for other Cisco programs. I just wanted to keep the archive as it as,extract what the package needs, and get rid of the rest.

About the /usr/lib directory, you're right. I'll modify the script later today.

RezaT4795:
@sbo-bot: build network/cisco-secure-client-vpn

willysr:
merged

PR #16430, opened 2026-07-05
development/claude-code: Updated for version 2.1.201

NicholasBHubbard:
`sbopkglint`'s  `20-arch` test fails. This is expected, with reasoning described in PR #16002.


cwilling:
Merged

PR #16549, opened 2026-07-14
academic/root: Update for version 6.36.14

arabusov:
The latest ROOT release requires at least python 3.10, while we have only 3.9 in the base system. Also, the 6.36 branch supposed to be LTS, so let's stay with 6.36 for Slackware 15.0.


cwilling:
@sbo-bot: build academic/root

cwilling:
Merged

PR #16553, opened 2026-07-14
network/mullvad-browser: Updated for version 15.0.18

antonioleal:
network/mullvad-browser: Updated for version 15.0.18


antonioleal:
@sbo-bot: build network/mullvad-browser

cwilling:
Merged

@antonioleal
Although approved and merged for next public update, it feels untidy to be using /usr/lib in a package only for x86_64.
For the next version update, could you consider also moving contents of /usr/lib to /usr/lib64?
It is easy to do - by adding the following lines just before the existing `mkdir -p $PKG` line. A quick runtime test at home showed no problems and leaves me with a cleaner system.

```
mv $TMP/$PRGNAM/usr/lib $TMP/$PRGNAM/usr/lib64
(cd $TMP/$PRGNAM/usr/bin && rm mullvad-browser ; ln -s ../lib64/mullvad-browser/start-mullvad-browser mullvad-browser)
sed -i -e 's:/lib/:/lib64/:' $TMP/$PRGNAM/usr/share/applications/mullvad-browser.desktop
```

What do you think?

PR #16652, opened 2026-07-20
desktop/yaru-icon-theme: Update for 26.04.5.1ubuntu

isaackwy:
The latest version of yaru-icon-theme is actually 26.10.1.

However, later versions of yaru have removed the GTK2 theme.
This is not directly related to yaru-icon-theme itself. Rather, if I were to upload yaru-gtk-theme, I would rather have yaru-icon-theme and yaru-gtk-theme be at the same version.


isaackwy:
@sbo-bot: build desktop/yaru-icon-theme

cwilling:
Merged

PR #16728, opened 2026-07-25
misc/fcitx5-mozc: New maintainer.

pghvlaans:
I'd push an update, but it would require `qt6`. That's fine for -current, but I'm not sure the new version is a big enough improvement to justify making everyone who wants this do the extra dependency builds.


willysr:
merged

PR #16764, opened 2026-07-26
games/fs-uae-arcade: Removed (use fs-uae-launcher --arcade)

nomnombtc:
This orphaned build is outdated and not needed anymore so it could be removed.
Arcade is included in fs-uae-launcher. Users can run `fs-uae-launcher --arcade`


cwilling:
@willysr 
What's the correct process for removal?
- when I fetch the PR to sign off, the games/fs-uae-arcade directory no longer exists so `git ci --amend games/fs-uae-arcade` fails so I can't sign off. Should I just add the PR as it is to my branch without sign off?
- should I remove it from the DB now or will you do that at global public update?

willysr:
just remove it from git for now
i will handle the DB during public update later on

willysr:
merged

PR #16790, opened 2026-07-27
misc/es40: Updated for version 0.76

antonioleal:
misc/es40: Updated for version 0.76


antonioleal:
@sbo-bot: build misc/es40

PR #16818, opened 2026-07-29
Added (Linux userspace driver for Valve's Steam Deck hardware.)

khronosschoty:


willysr:
@khronosschoty this should be added via submission form since it's a new script

PR #16897, opened 2026-08-01
perl/perl-Math-BigInt: Updated for version 2.005003.

pyllyukko:


pyllyukko:
@sbo-bot: build perl/perl-Math-BigInt

willysr:
merged

PR #16899, opened 2026-08-01
perl/perl-Mail-DMARC: Updated for version 2.20260724.

pyllyukko:


pyllyukko:
@sbo-bot: build perl/perl-Mail-DMARC

pyllyukko:
Right. So this requires newer Perl than we have in Slackware :/

From the ChangeLog:

```
### 2.20260724

- add <prefix>/etc/mail to search path
- warn when mail-dmarc.ini not found
- chore: bump minimum supported Perl to 5.40.2
```

PR #16916, opened 2026-08-03
development/rust-opt: Add aarch64.

pghvlaans:
Like it says in the README, I won't be doing the full reverse dependency checks for aarch64 just now (though I'll check one just to make sure it works). My plan is to stop ARM testing and pick that up after the repo turns over for the new Slackware version.

@willysr, there aren't any changes to the actual script, so I assume the reverse dependency tests aren't needed this time?

### DIFF
--- a/development/rust-opt/rust-opt.info
+++ b/development/rust-opt/rust-opt.info
@@ -5,8 +5,10 @@ DOWNLOAD="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-i686-unknown-
           https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-arm-unknown-linux-gnueabihf.tar.gz"
 MD5SUM="0c19ca3c1fa3701b0a64ce44ae3e7f02 \
         9d72f03340614775916fa291f8c30252"
-DOWNLOAD_x86_64="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-x86_64-unknown-linux-gnu.tar.gz"
-MD5SUM_x86_64="bbabf63ac45c56b9522164200f92353a"
+DOWNLOAD_x86_64="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-x86_64-unknown-linux-gnu.tar.gz \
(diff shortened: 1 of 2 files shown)


pghvlaans:
@sbo-bot: build development/rust-opt

development/rust-opt/rust-opt.info
```diff
@@ -5,8 +5,10 @@ DOWNLOAD="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-i686-unknown-
           https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-arm-unknown-linux-gnueabihf.tar.gz"
 MD5SUM="0c19ca3c1fa3701b0a64ce44ae3e7f02 \
         9d72f03340614775916fa291f8c30252"
-DOWNLOAD_x86_64="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-x86_64-unknown-linux-gnu.tar.gz"
-MD5SUM_x86_64="bbabf63ac45c56b9522164200f92353a"
+DOWNLOAD_x86_64="https://static.rust-lang.org/dist/2026-07-16/rust-1.97.1-x86_64-unknown-linux-gnu.tar.gz \
```
aclemons, line 8: Thank you 🙂 - now I don't have to carry this patch forward on my fork

aclemons:
> @willysr, there aren't any changes to the actual script, so I assume the reverse dependency tests aren't needed this time?

It's fine.

I haven't setup the CI scripts to work with -current since it's still constantly changing. So there is no way to trigger them at the moment.

development/rust-opt/rust-opt.info
pghvlaans: Thank Lockywolf, who sent the request. I was going to hold off for 15.1, but this should be fine.

pghvlaans: The trick is that I only have time to run either ARM or aarch64 tests, not both. It's automated on my end, but there are so many reverse dependencies now that doing ARM alone takes well over ten hours, maybe closer to 24.

willysr:
merged

PR #17040, opened 2026-08-07
desktop/anki: Reverted to version 24.11 to allow building on stable

chexmix:
anki versions after 24.11 require glibc > 2.35 to build, and Slackware 15.0 stable has glibc 2.33. 

I altered the anki.info and anki.SlackBuild files accordingly, and added a line (thanks, B. Watson!) to the latter to address an error with the anki.desktop file when I ran sbopkglint against the package.

I left maintainer info alone. If needed I can adopt this package, since I use it.


cwilling:
@sbo-bot: build desktop/anki

cwilling:
Merged

PR #17046, opened 2026-08-08
python/secretstorage: Removed (renamed to python3-secretstorage)

isaackwy:
I already have uploaded python3-secretstorage (and now taken over maintenance from Dimitris - he was the maintainer of secretstorage beforehand.)


willysr:
merged

PR #17047, opened 2026-08-08
python/python3-secretstorage: Fix README

isaackwy:
The README initially didn't specify which version is actually the last available one on Slackware 15.0.
This PR fixes this.


willysr:
merged

PR #17063, opened 2026-08-08
libraries/avr-libc: Updated for version 2.3.2.

newHeiko:
Also included a proper license in the SlackBuild.

Is that OK after taking over maintenance?


newHeiko:
@sbo-bot: build libraries/avr-libc

newHeiko:
This package traditionally puts its files into /usr/avr... I'd love to keep them there, but if you insist, I'll try and move them out... what would be the correct place? 

Heiko

willysr:
merged

PR #17096, opened 2026-08-10
network/proton-vpn-cli: Edit README (Add usage instructions)

isaackwy:
proton-vpn-cli is now at 1.0.3, but I have not updated it just yet.
It requires python3-proton-vpn-api-core to be upgraded to v5.5.6 (which is really difficult by itself.)

I also haven't been able to log in to Proton using the command `protonvpn signin username@email.address`.


willysr:
This is not what README is all about
they should look for documentation upstream about usage

PR #17119, opened 2026-08-12
development/antigravity2: please remove this package

antonioleal:
Please remove this package because it has been renamed to antigravity.

google has change the program names, so now:

antigravity becomes antigravity-ide and

antigravity2 becomes antigravity

This realigns SlackBuilds.org names with the ones used upstream


willysr:
merged

PR #17132, opened 2026-08-12
network/proton-vpn-cli: Removed (unable to login, unable to use)

isaackwy:
NOTE: proton-vpn-cli itself starts up successfully. Please see below for what the failure is.

I tried logging in to proton-vpn-cli on a Slackware virtual machine. I failed to do so.
I also tried using proton-vpn-gtk-app (an older version at 4.13.1) - the login failed as well.

proton-vpn-cli requires logging in to a Proton account to connect to a VPN.
Without logging in, I cannot use proton-vpn-cli.

Please keep all the other dependencies of proton-vpn-cli on SlackBuilds.org. When I return to using Slackware as my daily driver, I'll need them (ex. next Slackware release, I'll try to see if proton-vpn-gtk-app works.)


willysr:
merged

PR #17190, opened 2026-08-14
python/python3-construct: New Maintainer.

nomnombtc:
Hello, I can take this SlackBuild if no one else did already. (See Isaac Yu mailinglist post)


willysr:
merged

PR #17060, opened 2026-08-08
desktop/claude-desktop: Updated for version 1.25927.0

earies:

PR #17200, opened 2026-08-15
desktop/claude-desktop: Updated for version 1.30096.1

earies:


earies:
@sbo-bot: build desktop/claude-desktop

cwilling:
@sbo-bot: build desktop/claude-desktop

cwilling:
Please note sbopkglint message:
```
--- ERR: ELF binaries/libraries with bad glibc version:
usr/lib64/claude-desktop/resources/app.asar.unpacked/node_modules/@ant/claude-native/claude-native-binding.node
usr/lib64/claude-desktop/resources/chrome-native-host
usr/lib64/claude-desktop/resources/virtiofsd
FAILED
```

willysr:
@earies this is not working on stable since it requires newer GLIBC

earies:
@willysr @cwilling - looking at these binaries that were flagged also have the same min glibc version (2.34) as https://github.com/SlackBuildsOrg/slackbuilds/pull/17060 however there is no CI failure there (I would have expected failure there as well) - in fact I've walked back to the intial submission and all carry the same min glibc version

Was something adjusted on the runner side since for this to be caught now?

As far as the offending binaries, quick dig looks like one specific is mandatory for core functionality, closed sources, cannot drop so we're held captive here on `claude-desktop` as a whole on stable it seems

cwilling:
@earies  There have recently been so many updates of binary packages that require a "bad" glibc, that sbopkglint was modified to catch such updates.

The previous version you mention passed before that updated check. I just ran the claude-desktop version that had been approved through our CI again (see #17280) and it is now  flagged with the same problem.

As you say, this means many binary package versions will be restricted on -stable.

willysr:
yeah, the new sbo-maintainers-tools is updated last week by Urchlay to add glibc detection because we found some scripts were tested under -current and continue to build normally in CI until someone reported them to be broken under stable during runtime

PR #17222, opened 2026-08-17
Eliminate sbolint LIBDIRSUFFIX and SLKCFLAGS warnings

duncan-roe:

PR #17225, opened 2026-08-17
ruby/tty-screen: orphan

Tonus1:
Dep for already orphaned package


cwilling:
Duncan Roe may take over this SlackBuild (#17222)

cwilling:
Merged
- also removed LIBDIRSUFFIX & SLKCFLAGS (as per Duncan's #17222) in case someone picks this up

PR #17233, opened 2026-08-17
office/xpad: New Maintainer

antonioleal:
office/xpad: New Maintainer


antonioleal:
@sbo-bot: build office/xpad

cwilling:
@antonioleal  - I see you've added gtksourceview3 to REQUIRES. Is this intentional? Perhaps for builds on  -current (which has gtksourceview4 built in)?

antonioleal:
@cwilling I tried on -current and had to install gtksourceview3 in order to build.

willysr:
please try in -stable, SBo only supports -stable

cwilling:
It builds OK in -stable - looks like CI ignores a REQUIRES entry that doesn't exist.

Same as my local system: any unrecognized dependency is ignored and building proceeds. There is no gtksourceview3 in SBo  so it is ignored and the build succeeds because there exists a system gtksourceview3.

For -current, ponce's repo has a SlackBuild for gtksourceview3 and using it in REQUIRES is useful to override the system gtksourceview4 which I presume is unsuitable for xpad.

The question is: should we accept something that is nominally wrong but still works?
(well, it works in two different build systems that we've tried  - maybe not all build systems)

cwilling:
One answer to the question above is to be strict for SBo (remove non-existent gtksourceview3 from REQUIRES). Then ponce will have to add it in again in his repo to make it work in -current.

willysr:
@antonioleal try removing gtksourcevie3 and see if it builds
if it does, then yeah we can accept it @cwilling

antonioleal:
I run several -stables and one -current.

It does not build with gtksourceview4 on -current. gtksourceview3 is required.

I will amend the PR to remove gtksourceview3, Matteo can add it later on his -current repo.

I included only because no harm came to -stable by having it...

antonioleal:
@sbo-bot: build office/xpad

cwilling:
Merged

PR #17239, opened 2026-08-18
Eliminate sbolint LIBDIRSUFFIX and SLKCFLAGS warnings

duncan-roe:


cwilling:
@duncan-roe The email thread where this was discussed ends (more or less) with:
```
The best time to do this would be during 'development cycle',
when a new Slackware stable has just been released, and the repo
is closed while we port all the builds over (fix compile errors,
remove builds that were included in the new Slackware, etc).
```
i.e. not yet

duncan-roe:
> i.e. not yet

I was planning to respond on that thread once I'd finished testing the way I've scripted it (to be run by cron, sending a constant trickle). Testing might generate less than half a dozen more PRs - would that be OK?
Last test would be running as root, so I know it can go in root's cron.hourly. (That's about the extent of my knowledge of cron). I'd only actually put it in cron.hourly with agreement from you guys.

duncan-roe:
What triggers the bot to run the other tests?

cwilling:
> What triggers the bot to run the other tests?

The tests are run whenever someone submits an update (new version etc.). For the updates that I process, I usually remove LIBDIRSUFFIX and SLKCFLAGS if they've been flagged as useless. This means the maintainer will be aware of the change because I mention them when I merge the update.

Such a removal is a quick thing to add if I'm already processing an update; not so quick otherwise. That's why it was recommended to leave until the next development cycle.

@willysr  Willy do you have any opinion about these non-maintainer PRs to remove unused LIBDIRSUFFIX and SLKCFLAGS ?

willysr:
as long as it doesn't change anything else and the build passed our CI test, it should be fine

cwilling:
Merged

PR #17294, opened 2026-08-19
network/nordvpn: Updated for version 5.3.0.

RezaT4795:


RezaT4795:
@sbo-bot: build network/nordvpn

willysr:
@RezaT4795 this requires newer GLIBC

cwilling (commented):
I see the same glibc problem.

I also just built the previous version 5.2.0 (the existing version in SBo) and that survives sbopkglint OK

willysr:
yes, i checked the current version in repo is safe
i think it started to bump since 5.3.0, so i guess it's rejected then

PR #17295, opened 2026-08-19
network/nordvpn-gui: Updated for version 5.3.0.

RezaT4795:


RezaT4795:
@sbo-bot: build network/nordvpn-gui

willysr:
this is pending waiting for #17294

willysr:
@RezaT4795 will this work with nordvpn 5.2.0 ?
since we can't accept 5.3.0 due to GLIBC issue

RezaT4795:
> @RezaT4795 will this work with nordvpn 5.2.0 ?
> 
> since we can't accept 5.3.0 due to GLIBC issue

No. The versions should be matching. For now decline both of them. 
We'll probably have these for Slackware 15.1

PR #17320, opened 2026-08-20
academic/hp-prime: Moved src to sourceforge.net

antonioleal:
academic/hp-prime: Moved src to sourceforge.net


antonioleal:
@sbo-bot: build academic/hp-prime

PR #17334, opened 2026-08-21
development/sbt: Updated for version 2.0.7

perrin4869:


perrin4869:
@sbo-bot: build development/sbt

perrin4869:
I think the glibc check is more recent and is probably broken in the previous versions of sbt in slackware 15 as well, I don't see the error in slackware-current

willysr:
yes, it's a new detection since some people push changes that requires newer GLIBC

willysr:
@perrin4869 did you test this under -current?

perrin4869:
Yes, I rely on the ci here for 15.0

willysr:
that's a big no no
you should at least test it under VM for SBo

perrin4869:
sorry I had been complacent, I actually had a slackware container to run these tests but thought it was more practical to use the CI
I can try to find a version of this package that works with older glibc and downgrade or we can keep this one, since the main functionality of the package is not in the broken binary, but a jar file run under the jvm

perrin4869:
ok, I managed to find the last version that had `/usr/share/sbt/bin/sbtn-x86_64-pc-linux` built with a slackware 15.0 compatible glibc, it was 1.11.4. We can downgrade to that or keep on latest I imagine

willysr:
please switch to 1.11.4 and make a note in the README that if people want to upgrade to newer version, they need to upgrade GLIBC

PR #17354, opened 2026-08-21
network/wifi-radar: Fix Homepage in slack-desc; s/http/https in info

duncan-roe:


duncan-roe:
There should be 3 changed files but there's only 1

PR #17379, opened 2026-08-24
Fix slack-desc parentheses

pghvlaans:
About ten years ago, a number of `slack-desc` files had the first line cut off at the first right parenthesis. Restore any trailing text and missing parentheses so the descriptions show up properly in `SLACKBUILDS.TXT`, `sbopkg`, `sbotools`, etc.


willysr:
merged

PR #17449, opened 2026-08-28
system/elasticsearch: Orphaned.

nick8325:
I thought I would take over elasticsearch, but I've stopped using it and it looks annoying to package, so I'm dropping it again. Sorry!

The current SlackBuild is ~10 years out of date so probably any new maintainer would have to replace a lot of it.


cwilling:
Merged with change:
- removed unused LIBDIRSUFFIX & SLKCFLAGS

PR #17514, opened 2026-09-01
libraries/htslib: Updated to version 1.24

brobr:


cwilling:
@sbo-bot: build libraries/htslib

cwilling:
@brobr  - notice sbopkglint comment `___ note: usr/lib64 has both libhts.so and libhts.a`. We usually only package .a files when they're known to be needed for something. Please advise whether both .so & .a libraries are needed.

brobr:
No idea whether the .a one is used; how can I tell, via ldd? 

bash-5.3$ ldd /usr/bin/samtools
	linux-vdso.so.1 (0x00007fc8686cf000)
	libhts.so.3 => /usr/lib64/libhts.so.3 (0x00007fc868578000)
	libncursesw.so.6 => /lib64/libncursesw.so.6 (0x00007fc86853c000)
	libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fc868507000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fc86840a000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fc8683f0000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fc868000000)
	libdeflate.so.0 => /usr/lib64/libdeflate.so.0 (0x00007fc8683d6000)
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007fc8683a6000)
	libbz2.so.1 => /lib64/libbz2.so.1 (0x00007fc868393000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc8686d1000)


 Are the *.la files not the problematic ones...
Anyway will check and then edit the slackbuilfd; thanks for noticing; have not changed the script for years..

cwilling:
@brobr  Only the  samtools and bcftools packages depend on htslib. To test whether they make any use of libhts.a (assuming you have a local installation of these three and regardless of the versions you have), try removing (save it somewhere out of the way in case you really need it) the file /usr/lib64/libhts.a and then run whatever you would normally run from the samtools and bcftools packages e.g. /usr/bin/samtools. Do they still work OK? My guess is that they'll still work OK and, if so, it'll be safe to remove libhts.a during packaging.

cwilling:
@sbo-bot: build libraries/htslib

cwilling:
@sbo-bot: build libraries/htslib

cwilling:
@brobr  all good now. I made one small change to satisfy the `--- ERR: ELF binaries/libraries should be executable:` complaint.

BTW you can always test updates yourself before submission with `sbolint` and `sbopkglint` from the system/sbo-maintainer-tools SlackBuild at SBo.

cwilling:
Merged

brobr:
Hi Chris, thanks again; the problem is that even when run locally, an error like:

      --- ERR: ELF binaries/libraries should be executable:
      -rw-r--r-- 1 root root 1075096 Sep  2 10:39 usr/lib64/libhts.so.1.24
      FAILED

would not help me to come to the kind of a solution you've provided. 

Why did it work when I was testing it (by compiling samtools and bcftools, and that went fine); is it because I tend to run Slackbuilds as `sh prgnam.Slackbuild`? 

So, is this now a common addition (if so my and SBorg's autotools.template is outdated)? Or only specific to htslib?

       find $PKG -print0 | xargs -0 file | grep -e "executable" -e "shared object" | grep ELF \
       | cut -f 1 -d : | xargs chmod 0755 2> /dev/null || true

Sorry, too many questions, 
feel free to ignore

cwilling:
@brobr  The fact that the ELF executables are not executable is not a show stopper - everything should still work but it's traditional that they are. In the unusual event that something depending on them did not work, you'd want to be able to assume it wouldn't  be caused by the libraries not being executable.

Most build systems will do the right thing and create them the right way (executable) so we don't apply that fix by default. On packages with very large file structures (LibreOffice, CUDA toolkits) the searching can take a while so we don't want it done by default. So, no it's not a common addition - just when sbopkglint finds a problem.

PR #17515, opened 2026-09-01
network/onedrive: Updated to version 2.5.11

brobr:

PR #17523, opened 2026-09-02
development/dmd: Protect against exported BUILD

cwilling:
This is from belka's update branch to test whether it's good enough to build as  requirement for #17515 (currently FTB because of dmd failure)


cwilling:
@sbo-bot: build development/dmd

PR #17595, opened 2026-09-07
misc/exploitdb: Updated for version 2026_09_04.

danixland:
Test-built successfully on both -current and 15.0, sbolint clean.

The sbopkglint 15-noarch test reports ELF files under usr/share/exploitdb/bin-sploits (190-3.linux, 35081.bin). These are known exploit sample binaries shipped by the upstream exploitdb-bin-sploits secondary source, which stays pinned at 2022-11-22 and is unchanged by this update. The finding is pre-existing and not introduced by this version bump.


cwilling:
@sbo-bot: build misc/exploitdb

willysr:
merged

PR #17755, opened 2026-09-14
multimedia/devedeng: Updated for version 4.21.2

antonioleal:
multimedia/devedeng: Updated for version 4.21.2


antonioleal:
@sbo-bot: build multimedia/devedeng

cwilling:
Merged

@antonioleal Note sbopkglint output:
`Running test: 35-desktop...usr/share/applications/devede_ng.desktop: error: (will be fatal in the future): value "devedeng_icon.svg" for key "Icon" in group "Desktop Entry" is an icon name with an extension, but there should be no extension as described in the Icon Theme Specification if the value is not an absolute path`. Should be easy to fix for next update ...

PR #17809, opened 2026-09-18
games/openmw: Fix for version 0.51.0.

linrs:
fix gcc14 version mismatch


linrs:
@sbo-bot: build games/openmw

linrs:
@sbo-bot: build games/openmw

linrs:
@sbo-bot: build games/openmw

aclemons:
@willysr Could you unstage fa46ec222 and take this instead

cwilling:
Merged

@linrs  - just a suggestion - rather than setting `GCC14_VERSION=${GCC14_VERSION:-14.4.0}` and then having to do a lot of exports (like `export CC=gcc-"${GCC14_VERSION}"`), you could instead do `source /etc/profile.d/gcc14.sh`. That way any gcc14 version changes are taken care of automatically.

cwilling:
@linrs  - I just read that section of your script more carefully and see that you already source gcc14.sh and only do the multiple exports as a backup;  sorry about the useless advice!

However I have been thinking about some other mechanism to set up the environment correctly for gcc14. A problem with using /etc/profile.d/gcc14.sh is that once it is installed (for whatever reason), a new login or terminal will source that file and favour gcc14 over the default compiler regardless of whether it is actually wanted. For that reason I want to remove /etc/profile.d/gcc14.*sh file and have an equivalent somewhere else - maybe /usr/bin/gcc14-config or /etc/gcc14.conf or /etc/defaults/gcc14.

Whatever that location finally becomes, it would have to be explicitly sourced by a .SlackBuild for gcc14 to be enabled and wouldn't automatically be invoked when not wanted.

Do you have any thoughts or preferences about it?

linrs:
@cwilling The /etc/profile has 'if [ -x FILE ]' for all the scripts in '/etc/profile.d'. Since gcc14.sh has file mode 644, it wouldn't be sourced automatically.

cwilling:
@linrs Good point about the permissions. I'll relax then.

